PR #597, opened 2020-11-02
Titus frontend components alias and app source structure changes

kurdin:
- Added aliases into front end package to support easy/pretty imports. 
- Update to front end package components code to use aliases


mcdonnelldean:
@simoneb Opinion please on this, I am too far away from projects day to day. I know @mcollina Is not fond of this pattern but it would be good to get a third opinion to match our for and against :D

simoneb:
> @simoneb Opinion please on this, I am too far away from projects day to day. I know @mcollina Is not fond of this pattern but it would be good to get a third opinion to match our for and against :D

After looking at this implementation, I recommend that we don't implement this approach. I think Titus should not use by default react-app-rewired, especially because there is a better option out there. I would close this PR in favor of #644 , which uses a more standard approach at achieving the same.

simoneb:
Closed in favor of #644

PR #628, opened 2020-11-17
fix: avoid autoload on auth plugins in backend

simoneb:
because the index.js file conditionally registers them. Looks like something has changed in the behavior of fastify-autoload which recursively registers plugins in the plugins folder, whereas we don't want that for the authorization plugins, since its index.js entrypoint conditionally registers the auth plugin based on configuration.

This wasn't caught by automated tests possibly because no tests execute the whole server.

### DIFF
--- a/packages/titus-backend/lib/plugins/auth/index.js
+++ b/packages/titus-backend/lib/plugins/auth/index.js
@@ -9,7 +9,7 @@ const authProviders = {
 }
 
 async function auth(server, options) {
-  server.register(authProviders[options.auth.provider || 'auth0'], options)
+  server.register(authProviders[options.auth.provider], options)
 }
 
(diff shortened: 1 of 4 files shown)


packages/titus-backend/lib/plugins/auth/index.js
```diff
@@ -9,7 +9,7 @@ const authProviders = {
 }
 
 async function auth(server, options) {
-  server.register(authProviders[options.auth.provider || 'auth0'], options)
+  server.register(authProviders[options.auth.provider], options)
```
simoneb, line 12: no need to hardcode the default, it comes from the config

salmanm (approved):

mcollina (approved):
lgtm

PR #644, opened 2020-12-10
refactor: aliases

simoneb:
Introduces support for aliases in titus-frontend by using `jsconfig.json` in order to make non-relative imports relative to the root of the project.

Also restructures the project a bit to avoid too many files in the root and moving them to a lib subfolder.

Supersedes #597


kurdin:
@simoneb is just adding `jsconfig.json` enough to make aliases work for ts and babel? If so, yeah, that would be a better approach to make non-relative imports relative to the root of the project. I did not see `jsconfig.json` used before.

simoneb:
> @simoneb is just adding `jsconfig.json` enough to make aliases work for ts and babel? If so, yeah, that would be a better approach to make non-relative imports relative to the root of the project. I did not see `jsconfig.json` used before.

Yes, the code you see in this PR works. Look up jsconfig in Google, for TS you'd use tsconfig but the idea is the same

kurdin:
> > @simoneb is just adding `jsconfig.json` enough to make aliases work for ts and babel? If so, yeah, that would be a better approach to make non-relative imports relative to the root of the project. I did not see `jsconfig.json` used before.
> 
> Yes, the code you see in this PR works. Look up jsconfig in Google, for TS you'd use tsconfig but the idea is the same

Not sure if I understand what jsconfig.json does. Is it VSCode only thing ? Seems like it changes ts config but that about webpack ? Sorry I don't have time right now to test it out.

simoneb:
> > > @simoneb is just adding `jsconfig.json` enough to make aliases work for ts and babel? If so, yeah, that would be a better approach to make non-relative imports relative to the root of the project. I did not see `jsconfig.json` used before.
> > 
> > 
> > Yes, the code you see in this PR works. Look up jsconfig in Google, for TS you'd use tsconfig but the idea is the same
> 
> Not sure if I understand what jsconfig.json does. Is it VSCode only thing ? Seems like it changes ts config but that about webpack ? Sorry I don't have time right now to test it out but I can look later or tomorrow

https://create-react-app.dev/docs/importing-a-component/#absolute-imports

mcdonnelldean (approved):

PR #645, opened 2020-12-10
Fix/husky upgrade

### LINKED ISSUE #643: Broken husky hooks
I experienced 2 issues with the lint staged:

1. when committing changes to multiple packages, lint-staged with lerna exploded :boom: I had to skip the hooks to commit. I'll post the console output in a follow up comment. This probably requires the concurrency setting with lerna `lerna run lint:staged --concurrency 1`
1. husky v5 is not backwards compatible with v4 which was merged in #640 . See the [husky docs](https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v5). Seems the hooks and lint-staged are now disabled :cry:

codyzu:
fixes #643 

1. restores the husky hooks as per the [husky docs](https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v5) for v5.0.4 of husky that was automerged by renovate
1. sets the lerna concurrency to 1 for lint-staged (lint-staged would :boom: if there were staged changes in multiple packages, it cannot be run in parallel)


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
codyzu, line 1: This file is auto-created by husk. Seems it is required because husky makes a temp directory name `_`.

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname $0)/_/husky.sh"
```
simoneb, line 2: why do we need this? the docs don't mention it

package.json
```diff
@@ -33,11 +33,5 @@
...
     "lerna": "^3.22.1"
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "npm run lint:staged && npm run test:related",
-      "pre-push": "npm run lint:all && npm run test"
-    }
   }
-}
+}
```
simoneb, line 37: always keep a new line at the end of files

simoneb (changes requested):
See inline comments, plus:

- from the docs, HUSKY_SKIP_INSTALL is removed, and it's being used somewhere in the codebase. Please find a suitable replacement

.husky/pre-commit
codyzu: This is added automatically when running `npx husky add ...` to add a new hook.

Installing husky creates a "hidden" directory `_` that contains `husky.sh` that does some error handling and friendly error messages. Husky now gets triggered by git config, instead of previously adding shell scripts in the `.git/hooks`.

In the end, those lines are automatically added by husky. The actual `husky.sh` file is not intended to be checked in. They seem to have completely reworked their integration in version 5.

simoneb: understood, thanks. I was asking as this is not shown in the docs

simoneb:
@codyzu did you have a chance to look into https://github.com/nearform/titus/pull/645#pullrequestreview-549942787?

codyzu:
> @codyzu did you have a chance to look into [#645 (review)](https://github.com/nearform/titus/pull/645#pullrequestreview-549942787)?

@simoneb I read through the husky docs, there is no longer an env var to disable the install. Instead, it needs to be handled in the npm postinstall script (i.e. don't call `husky install` if it shouldn't be installed). This means some other means must be hardcoded into the postinstall script to detect if husky should be installed.

I don't know enough about mira to know if there are any existing env vars that could be used. Although that seems a bit out of place. It smells like a more generic solution should be found. Perhaps the mira buildspec should be refactored to not call `npm run postinstall` (P.S., why is that called explicitly???). This revealed more questions than answers and will require some investigation...

simoneb:
Superseded by #649

Issue #642, opened 2020-12-07
Confusing route prefixes


codyzu:
Titus routes are automagically added with [fastify-autoload](https://github.com/fastify/fastify-autoload). This happens [here](https://github.com/nearform/titus/blob/28682c1b11bd8958f78d23b79c2688325117aa50/packages/titus-backend/lib/server.js#L17). The fasitfy-autoload [`dirNameRoutePrefix` option](https://github.com/fastify/fastify-autoload#global-configuration) is not set. As noted in the documentation, this defaults to true, meaning fastify-autoload includes the directory name as the route prefix.

The plugins `healthcheck` and `user` are wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin).
The plugin `log` is not wrapped with fastify plugin.

As noted in the [fastify docs](https://www.fastify.io/docs/latest/Plugins/#route-prefixing-option), fastify-plugin is not compatible with the prefix (the prefix is not respected).

This results in the following working routes out of the box with a new Titus project:

* GET: /healthcheck
* GET: /user
* POST: /log/log :warning: **I assume this is not the intended behavior.**

As described above, the extra `log` path segment is because the log plugin route is not wrapped with fastify-plugin, meaning routes will respect the route prefix that is auto-magically added by fastify-autoload.

This feels confusing and should probably be simplified.

Either disable the `dirNameRoutePrefix` in fastify-autoload so that no matter if plugins are wrapped with fastify-autoload or not, the resulting route will be the same. Note that plugins loaded with fastify-autoload can override the prefix with some [options](https://github.com/fastify/fastify-autoload#plugin-configuration). Alternatively, this should be documented so that the resulting behavior is understandable.

What is the desired behavior?

simoneb:
I think this is an oversight. I would suggest to use fp for all the plugins in order to avoid unwanted prefixes and make them all consistent. Please also fix any request hitting the /log/log endpoint currently, if any.

codyzu:
This is probably a problem with the perfectionist in me... I'm not a big fan of auto-loading plugins.

The downside of using fp everywhere is that prefixes will no longer work. While this is ok for simple cases, it really eliminates some of the benefits of auto-loading where you could build a hierarchy with the directory structure (imagine a `v1` and `v2` directories). Using fp everywhere would force plugins to add the prefixes themselves. :thinking: 

[...]

PR #647, opened 2020-12-11
normalize routes

mcollina:
Closes https://github.com/nearform/titus/issues/642

### DIFF
--- a/packages/titus-backend/lib/routes/healthcheck/healthcheck.test.js
+++ b/packages/titus-backend/lib/routes/healthcheck/healthcheck.test.js
@@ -24,7 +24,7 @@ describe('health route', () => {
     server.pg.query.mockResolvedValue({ rowCount: 1 })
     const response = await server.inject({
       method: 'GET',
-      url: '/healthcheck'
+      url: '/'
     })
 
     expect(response.statusCode).toEqual(200)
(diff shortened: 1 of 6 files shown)


simoneb:
This change is not ok because it changes the routes. Without using fp, it will prefix the routes. 

Here's what routes are without this change (only the relevant part is included):

```
└── /
    ├── user (GET)
    │   └── list (GET)
```

And this is what routes are with this change:

```
└── /
    ├── user
    │   ├── list (GET)
    │   └── /user (GET)
```

See that is has prefixed the `/user` route with `user`, making it `/user/user`.

This is something that was mentioned in https://github.com/nearform/titus/issues/642.

Nevertheless, the good thing is that this route is probably unnecessary and in fact never called in the app (it's called in some commented code). Plus, this route implementation is even wrong because it doesn't check if the user is authenticated. If it's not, fastify hangs forever because the route handler is returning undefined:

```
Promise may not be fulfilled with 'undefined' when statusCode is not 204
```

mcollina:
This should be good now. I'm a bit skeptical of the fact that the CI passed.

simoneb:
There are test, as you can see because the build is broken now, but they test the plugin in isolation, hence they don't cover the auto prefixing behavior.

mcollina:
Anyway, this needs to be investigated as it shows a bad practice.

skywickenden:
I've normalised the routes. Including renaming the `health` folder to `healthcheck` 

After discussion with @simoneb  I'm going to include the removal of the `user` route in https://github.com/nearform/titus/pull/594 as it deals with a related issue and will result in conflicts

simoneb (approved):
LGTM

mcollina:
lgtm, land when you want

PR #649, opened 2020-12-15
fix: rollback husky and prevent it from happening again

### LINKED ISSUE #643: Broken husky hooks

simoneb:
Supersedes https://github.com/nearform/titus/pull/645/
Fixes #643 

Introduces two renovate options to prevent this from happening in the future:

- stability days which prevent bumping dependencies unless they've been released for at least 7 days
- pr creation which doesn't create PRs until the deps are ready to be updated (meaning that 7 days have elapsed)

Also, it limits lerna's concurrency when running lint:staged script to 1 as it breaks otherwise due to a lint-staged issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "db:up": "lerna run --stream db:up",
     "lint:all": "lerna run --parallel lint",
     "lint:fix": "lerna run --parallel lint:fix",
-    "lint:staged": "lerna run lint:staged",
+    "lint:staged": "lerna run lint:staged --concurrency 1",
     "postinstall": "lerna bootstrap",
     "start:all": "lerna run --parallel start",
     "start:docs": "docsify serve ./docs --port 4000",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "db:up": "lerna run --stream db:up",
     "lint:all": "lerna run --parallel lint",
     "lint:fix": "lerna run --parallel lint:fix",
-    "lint:staged": "lerna run lint:staged",
+    "lint:staged": "lerna run lint:staged --concurrency 1",
```
codyzu, line 21: where can we document why this is needed??? I guess indirectly it is mentioned in my initial issue, but little things like this seem important to be documented :thinking:

codyzu (approved):
LGTM

wondering how or where we can document the concurrency issue with lint-staged...

package.json
simoneb: I'll put this in the PR description

PR #651, opened 2020-12-16
Create an alias for storybook

skywickenden:
This fixes storybook file imports being broken due to https://github.com/nearform/titus/pull/644

Adds a custom webpack alias for `.storybook` so that storybook config files can be imported absolutely in stories.


simoneb (approved):
LGTM

Issue #635, opened 2020-11-27
Update to node 14


codyzu:
Node 14 is LTS but the `.nvmrc` file in Titus is node 12. Node 12 will be in maintenance LTS on 30 November 2020.

skywickenden:
Unfortunately [AWS lambdas don't support 14 yet](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) so the `titus-infra-aws-mira` example can't be updated. We should wait until it is available before we go ahead.

mcollina:
> We should wait until it is available before we go ahead.

Please let's start updating CI so that we start running all the tests and builds using v14 as well, e.g. https://github.com/nearform/titus/blob/679e6ea4fd586c0634441a06beb2f23fcedc3bde/.github/workflows/backend-ci.yml#L22.

simoneb:
closed by #652

codeflyer:
To solve this issue an update to Mira is required. 
There is a task already assigned to @nigelhanlon to update Mira to the latest CDK version.

simoneb:
@codeflyer @nigelhanlon is that issue linked to this one then?

codeflyer:
The link to the Mira Maintenance task:

https://github.com/nearform/mira-archive/issues/434

nigelhanlon:
Currently, there are a number of build errors when trying to use CDK v1.89.0 with Titus. Below is the output of `npm run build`.  I am attempting to resolve them so we can use the pending Mira update. 

```bash
➜  titus-infra-aws-mira git:(master) ✗ npm run build

> titus-infra-aws-mira@1.0.0 build
> tsc

cdk/api-gw-alb/index.ts:51:20 - error TS2345: Argument of type 'this' is not assignable to parameter of type 'Construct'.
  Type 'ApiGatewayAlb' is not assignable to type 'Construct'.
    Property 'onValidate' is protected but type 'Construct' is not a class derived from 'Construct'.

51       new Function(this, 'GetConfigFunction', {
                      ~~~~

cdk/api-gw-alb/index.ts:151:39 - error TS2345: Argument of type 'this' is not assignable to parameter of type 'Construct'.

151     const logGroupProd = new LogGroup(this, `ApiGWLogGroup${stage}`, {
                                          ~~~~

cdk/api-gw-nlb/index.ts:56:20 - error TS2345: Argument of type 'this' is not assignable to parameter of type 'Construct'.
[...]

PR #652, opened 2020-12-16
Chore: Node version to 14

skywickenden:
Update node version from 12 to 14 for https://github.com/nearform/titus/issues/635

As suggested by @mcollina I've added CI checks for 12 and 14 and left the aws example at 12.


mcollina (approved):
lgtm on green CI

simoneb (approved):
LGTM

Issue #657, opened 2021-01-05
Implement authorization via fastify-casbin


simoneb:
Please refer  to #594 as a starting point and implement something similar for auth0 and azure as well.

### Expectations

- [ ] the UI should show a button which performs a dummy action in the backend
- [ ] remove the unnecessary logic around listing and deleting users that's in #594, we don't want to overlap the authorization functionality and the interaction with the auth providers
- [ ] the dummy action in the backend should check if the authenticated user belongs to the "admin" group in the authentication provider (note that groups are called roles in auth0)
- [ ] if the authenticated user belongs to the admin group the action is allowed, otherwise an error 403 is returned and displayed in the UI
- [ ] the authorization check is done with fastify-casbin and corresponding casbin models and policies. different providers may need different models and policies
- [ ] in #594 you can see that with Cognito checking the association between a user and a group is easy because it's embedded in the JWT token generated by Cognito. This doesn't happen with auth0 and azure. Consider either using a different approach for auth0 and azure or using a unique approach for all 3, which would consist in loading the information about the association user<->group at runtime using the authentication provider APIs and using that for carrying out the authorization checks
- [ ] consider different approaches at writing the casbin policy/model. the way it's done in #594 consists of a single policy but policies could even be loaded at runtime and added programmatically to casbin

PR #663, opened 2021-01-08
feat: test casbin integration for auth providers 

admataz:
Makes a start at https://github.com/nearform/titus/issues/657
Extends work in existing branch at #594 

This PR sets up an endpoint in `titus-backend` for each of the three provided Authentication services that accepts a custom header for the current authenticated user's `idToken` from the authenticating service. 

This allows us to perform some faux-authorization and illustrate the use of Casbin by matching the users email address with a predefined value in `.env`

On the frontend - the Auth services' components `getUserData()` functions have been adjusted to include the `idToken` This token is used in the new `authz-check` component, which is essentially a button that triggers a request to the protected endpoint on the backend at `/authzcheck` where authorization is performed based on Casbin policies.


packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,58 @@
...
+        const { auth } = options
+        let isAdmin = false
+
+        // this is a sample to check the signed-in user's permission to access this resource via casbin policies based on configuration in the OAuth provider
+        // a real-world example would have more context around the user account and roles/scopes/permissions and policies
+
+        if (auth.provider === 'auth0') {
+          // depending on how this is managed - you could get associated user data from auth0 user management API like https://auth0.com/docs/api/management/v2#!/Users/get_user_roles
+          // or set this up to be included in the token via Auth0 rules https://auth0.com/docs/rules
+          isAdmin = true
```
admataz, line 30: always true for now - involves some set up in Auth0, or another way to associate the user with `admin`

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        if (auth.provider === 'cognito') {
+          // Cognito includes group information in the token, easy!
+          isAdmin = (request.user['cognito:groups'] || []).includes('admin')
+        }
+
+        if (auth.provider === 'azureAD') {
+          // sorry - haven't made this example yet
+          isAdmin = false
```
admataz, line 40: Azure needs some special TLC to get configured

packages/titus-frontend/src/index.js
```diff
@@ -11,12 +12,10 @@ import App from 'app'
 import { register as registerServiceWorker } from 'serviceWorker'
 
 render(<App />, document.getElementById('root'))
-
 /*
   Replace above `render` with the below, to enabled Azure AD authentication
-
   runWithAdal(
-    authContext,
+    getAuthContext(config),
```
admataz, line 18: the `authContext` was null - it needs to be fetched with the config

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,58 @@
+'use strict'
+
+const { Forbidden } = require('http-errors')
+const fp = require('fastify-plugin')
+
+async function authzChecks(server, options) {
```
simoneb, line 6: we should not apply conditional logic in this route depending on the provider. we have two options:

- if the route has a different implementation depending on the provider, then we should register the route conditionally as is already happening when each of the provider's plugins is registered conditionally
- otherwise we keep the implementation of the route the same but  move the conditional logic elsewhere. in other words, we should not see if(provider === "...") anywhere (or at least not in any new places in the codebase), as it's the wrong approach

docs/developers/packages/titus-backend/README.md
```diff
@@ -50,6 +57,19 @@ The `GET /healthcheck` endpoint is intended for your production cluster. Your ba
 it returns your application version and server timestamp. It also runs a dummy query against the database to ensure it's available.
 
 
+## Configure Authorization
+Titus provides authorization capabilities via Casbin, an access control system with configurable policies and models that allow you to control which resources in your application a user can access. 
+
+### Authz Check Routes
+Each of the Authentication service configurations included with Titus (Auth0, Cognito and AzureAD) provide the endpoint `/authzcheck` to illustrate the function of authorization with Casbin policies. These simple examples accept a standard JWT as an `idToken` as provided by the auth service and decode the token, extracting the user's email to match to any defined by the comma-separated value for `CHECK_AUTHZ_ADMIN_USERS` in `.env`
+
+This is a simple example to illustrate integration with Casbin. Your real-world app will require specific authz policies - Casbin supports a range of capabilities via _watchers_ and _adaptors_.
```
simoneb, line 66:
```suggestion
This is a simple example to illustrate integration with Casbin. Your real-world app will require specific authz policies - Casbin supports a range of capabilities via _watchers_ and _adapters_. 
```

docs/developers/packages/titus-backend/README.md
```diff
@@ -124,19 +144,26 @@ cognito.adminGetUser(params, (err, response) => {
...
-[custom attribute]: https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html#user-pool-settings-custom-attributes
-[validation instructions]: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html
-[Titus Database Manager]: /developers/packages/titus-db-manager/
 [node-postgres]: https://node-postgres.com/
+[Nodemon]: https://nodemon.io
+[Pino]: http://getpino.io
+[Prettier]: https://prettier.io
+[Standard]: https://standardjs.com
+[Titus Database Manager]: /developers/packages/titus-db-manager/
+[validation instructions]: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html
```
simoneb, line 168: do we really have these many new links in this file? it doesn't look like. In general, thought I appreciate you didn't introduce this approach, I prefer inline links because if you put them at the bottom you never know which ones are really used.

docs/developers/packages/titus-frontend/README.md
```diff
@@ -85,6 +85,12 @@ To enable it, do the following:
 1. Set `REACT_APP_AUTH_PROVIDER` to `AD` in the **`titus-frontend`** `.env` file.
 3. Change the `render` method in `src/index.js` to the commented out replacement that wraps it in `runWithAdal`.
 
+
+## Authorization example
+_Authentication_ (above) allows your app to verify the identity of your users, while _Authorization_ allows you to manage which resources in your app a user can access.  **`titus-frontend`** provides a an example authz request in `src/components/authz-check` that requests a privileged resource from the backend. 
+
+Refer to **`titus-backend`** [Configure Authorization] where policies specific to your application.
```
simoneb, line 92: did you intend to put a link to the backend config here? not sure if I'm doing anything wrong but the link doesn't seem to work for me.

packages/titus-backend/.env.sample
```diff
@@ -44,3 +44,7 @@ SECRETS_PG_PASS=PG_PASS
 #
 # SECRETS_STRATEGY=aws
 # SECRETS_PG_PASS=DB_PASSWORD
+
```
simoneb, line 47: nit: one too many new lines

packages/titus-backend/lib/plugins/auth/auth0/auth0-routes.js
```diff
@@ -90,6 +92,45 @@ async function authRoutes(server, options) {
         return user
       }
     })
+
+  // this is a sample to check the signed-in user's permission to access this resource via casbin policies based on configuration in the OAuth provider
+  // a real-world example would have more context around the user account and roles/scopes/permissions and policies
+  server.route({
+    method: 'GET',
```
simoneb, line 99: isn't this code exactly the same for all providers? then in that case we should not have one route for each provider but only one for all. the approach at registering the routes conditionally that I had suggested assumed that the route's code would be different per provider, but since the approach you took makes it the same, then there's no need to have multiple identical routes registered conditionally. just have a single one outside of the provider's folders that's always registered.

simoneb (changes requested):
The major requested change is around the duplication of the authz route. See inline comment but in summary, since this route is the same across the providers it doesn't need to be duplicated and registered conditionally.

There is one more thing I'd like to suggest, though not mandatory but if you have time. Seeing the approach you've decided to take in implementing this new approach, instead of reading the environment variable containing the emails of the users with admin privileges in the route, you could use casbin's grouping policies instead.

It would be more casbin-idiomatic in a sense. Here's how it would work:

- in the policy, you wouldn't need an expression anymore, but you would have a policy that allows admin access.
- when the casbin plugin is registered, you could load the environment variable with those emails and add grouping policies for each of them, basically saying that this email belongs to the admin group
- this would also require a change (a simplification in a sense) to the casbin model
- the route now would only check if the email it decodes from the header has access, without  having to construct an { isAdmin } object.

docs/developers/packages/titus-backend/README.md
admataz: I agree that inline links are better - I just sorted the list here to make it easier to manage

admataz:
Thank you @simoneb 
I've refactored the Azure authentication and made it a decorator so the authzcheck can be shared between the providers. Tests are currently broken as a result - which I will get to next. 

I agree - a more casbin approach is better - and I like your suggestion. I'll take a look at that next

simoneb (commented):
This overall LGTM, there's only one code change I'd like to see before we proceed with this. Can you please move the authzcheck route as a standalone module in the routes folder rather than registering it in a plugin? the plugin can still contain the hook to add the policies to casbin, but the route registration should be more conceptually correct that it lives in the routes directory.

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,52 @@
...
+
+async function authRoutes(server, options) {
+  // add users to the admin role
+  server.addHook('onReady', async function () {
+    const adminUsers = process.env.CHECK_AUTHZ_ADMIN_USERS || ''
+    const policies = adminUsers
+      .split(',')
+      .map(u => server.casbin.addRoleForUser(u, 'role_admin'))
+    await Promise.all(policies)
+  })
```
mcollina, line 15: Why did you put this in `onReady` and not in line? If you need to specifify a plugin dependency with fastify-autoload you can use the plugin dependencies: https://github.com/fastify/fastify-autoload#plugin-configuration:

```js
// plugins/plugin-a.js
const fp = require('fastify-plugin')

function plugin (fastify, opts, next) {
  // plugin a
}

module.exports = fp(plugin, {
  name: 'plugin-a',
  dependencies: ['plugin-b']
})

// plugins/plugin-b.js
function plugin (fastify, opts, next) {
  // plugin b
}

module.exports = fp(plugin, {
  name: 'plugin-b'
})
```

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,52 @@
...
+        {
+          apiKey: []
+        }
+      ]
+    },
+    onRequest: [
+      server.authenticate,
+      async (request, res) => {
+        const authzId = request.headers['x-authz-id']
+        const authzUserData = jwt.decode(authzId)
```
mcollina, line 34: didn't we have the user already decoded somewhere? We should try avoiding decoding twice. I would prefer verifying vs decoding (if possible).

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,52 @@
...
+        // we added the user to role with admin access above in the server.onReady hook
+        // Here's where the actual policy check goes ahead - based on the user's email
+        if (
+          !(await server.casbin.enforce(authzUserData.email, 'admin', 'access'))
+        ) {
+          throw new Forbidden('Cannot access admin')
+        }
+      }
```
mcollina, line 42: Verifying a role should really be a generic hook (similar to `server.authenticate`), e.g. `server.enforceRoleHook('admin')`.

packages/titus-backend/lib/plugins/auth/authzcheck.js
```diff
@@ -0,0 +1,52 @@
...
+    ],
+    handler: async req => {
+      return {
+        isAdmin: true,
+        user: req.user
+      }
+    }
+  })
```
mcollina, line 50: I would _not_ include a route inside something in `/plugins`, it should be in `/routes`.

mcollina (changes requested):

packages/titus-backend/lib/plugins/auth/authzcheck.js
simoneb: I believe, but I'll let Adam explain, that this is a different token than the one we already used. What we used is an access token and this one is an id token.

packages/titus-backend/lib/plugins/auth/authzcheck.js
simoneb: Yep similar to what I commented. Let's move this route registration as a standalone module in /routes.

packages/titus-backend/lib/plugins/auth/authzcheck.js
admataz: yes, defining the dependencies is much better. - Thanks

packages/titus-backend/lib/plugins/auth/authzcheck.js
admataz: Exactly - the `accessToken` doesn't always contain any useful information for assigning roles - but all the auth providers all also define `idToken` with more user info like email

packages/titus-backend/lib/plugins/auth/authzcheck.js
admataz: yes, makes sense, thank you.

packages/titus-backend/lib/plugins/auth/authzcheck.js
admataz: done - moved. Much better :)

admataz:
@simoneb @mcollina  - thank you both for the feedback - I think I've addressed all the issues raised and it's much improved.

simoneb (approved):
LGTM

PR #664, opened 2021-01-08
chore: remove unused packages

### LINKED ISSUE #656: Remove unnecessary packages
It seems that some packages are no longer used in  titus-frontend:

```
pretty-quick  
puppeteer
react-is
react-test-renderer
typescript
serve-handler
eslint-plugin-import
eslint-plugin-promise
```

make sure they are indeed not necessary (some may be used even if not referenced directly, as it may be peer dependencies of other packages), and remove those which are not used

RudyRed:
closes #656 

`typescript`, `eslint-plugin-promise`, `eslint-plugin-import` are all required by the linter

`react-is` is the only peer dependancy, however its not imported in the code and not needed in the `package.json`

### DIFF
--- a/packages/titus-backend/package-lock.json
+++ b/packages/titus-backend/package-lock.json
@@ -11565,7 +11565,8 @@
 		"growly": {
 			"version": "1.3.0",
 			"resolved": "https://registry.npmjs.org/growly/-/growly-1.3.0.tgz",
-			"integrity": "sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE="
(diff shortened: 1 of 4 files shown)


simoneb:
Can you double check react-test-renderer? I see the tests are succeeding but I am fairly sure it should be a peer dependency of react testing library so I'm somewhat surprised

RudyRed:
Double checked and it doesnt seem to be. I also out [npm](https://www.npmjs.com/package/@testing-library/react) and also looked at the [repo](https://github.com/testing-library/react-testing-library/blob/master/package.json), cloned it downloaded it and looked through and it doesnt seem to be anywhere there.

simoneb (approved):

PR #675, opened 2021-01-14
fix: gcp deployment

simoneb:


mcollina (approved):
lgtm

simoneb:
> lgtm

I'm still experimenting with this. I'll merge once the gcp builds are green

simoneb:
@mcollina this change makes all the gcp builds green. whether that means that the app is running fine I can't be sure but it's a first step at least.

PR #700, opened 2021-02-03
Fix #684: Use under-pressure on all routes and implement the check with healthCheck option

### LINKED ISSUE #684: refactor healthcheck
under-pressure is currently set up only for the healthcheck while it should be set up for all routes. Morever, it should use https://github.com/fastify/under-pressure#status-route and https://github.com/fastify/under-pressure#custom-health-checks instead of building its own route.

p16:
This PR fixes #684 

In this PR we are using the latest update to [`under-pressure`](https://github.com/fastify/under-pressure/pull/84) to replace the `/healthcheck` implementation with the one from `under-pressure`.

To do so we are:
- adding the response schema configuration and the custom url to the `config.underPressure` object
- extract the old `/healthcheck` implementation into a function that we then pass in into the `healthCheck` option of the `under-pressure` plugin


packages/titus-backend/lib/config/index.js
```diff
@@ -56,7 +57,24 @@ module.exports = {
...
+          type: 'object',
+          properties: {
+            eventLoopDelay: { type: 'string' },
+            rssBytes: { type: 'number' },
+            heapUsed: { type: 'number' }
+          }
+        },
+        db: { type: 'string' }
+      }
+    }
```
mcollina, line 76: please use fluent-json-schema here

packages/titus-backend/lib/config/index.js
```diff
@@ -56,7 +57,24 @@ module.exports = {
     poolSize: 10,
     idleTimeoutMillis: 30000
   },
-  underPressure: {},
+  underPressure: {
```
mcollina, line 60: while we are at it, I would also add some basic config for under-pressure:

```js
{
  maxHeapUsedBytes: 768 * 1024 * 1024, // arbitrary, 768 MB of RAM
  maxRssBytes: 1024 * 1024 * 1024, // arbitrary, 1 GB of RAM
  maxEventLoopUtilization: 0.98
}
```

mcollina (changes requested):

packages/titus-backend/lib/config/index.js
p16: @mcollina to clarify

Are you asking to update the name of the package from `fluent-schema` to `fluent-json-schema`?
Or are you asking to create a schema of the configuration?

packages/titus-backend/lib/config/index.js
p16: done and pushed, but it's not showing up yet 🤔

packages/titus-backend/lib/config/index.js
simoneb: I believe he's asking to use the fluent-json-schema library to define this json schema

packages/titus-backend/lib/plugins/healthcheck/index.js
```diff
@@ -0,0 +1,41 @@
...
+      healthCheck: async () => runCheck(parent)
+    }
+  })
+
+  next()
+}
+
+module.exports = fp(healthCheck, {
+  name: 'healthCheck',
+  dependencies: ['pg']
```
simoneb, line 40: this dependency doesn't look right. the name of the plugin is 'fastify-postgres'. please make sure you test this by running the app, this may pass the test but I think it breaks the app

packages/titus-backend/lib/plugins/healthcheck/index.js
```diff
@@ -0,0 +1,41 @@
...
+    memoryUsage: server.memoryUsage && server.memoryUsage(),
+    db: dbRes && dbRes.rowCount === 1 ? 'ok' : 'fail'
+  }
+}
+
+async function healthCheck(server, options, next) {
+  server.register(require('under-pressure'), parent => {
+    return {
+      ...options.underPressure,
+      healthCheck: async () => runCheck(parent)
```
simoneb, line 31: this function doesn't need to be async

packages/titus-backend/lib/plugins/healthcheck/index.js
p16: I'm requiring our own plugin, named `pg`

p16: I can require the generic one though

packages/titus-backend/lib/config/index.js
simoneb: @p16 here's an example

```
routeResponseSchemaOps: S.object().prop('version', S.string()).prop('serverTimestamp', ...
```

packages/titus-backend/lib/plugins/healthcheck/index.js
simoneb: ah sorry I missed that. I find it pointless to have that plugin but I understand now. it's fine

packages/titus-backend/lib/config/index.js
mcollina: > Are you asking to update the name of the package from fluent-schema to fluent-json-schema?


Also

p16: @mcollina @simoneb name updated and schema created

simoneb (approved):
LGTM. can you do a sanity check run by running the app locally and trying to trigger a failure of the healthcheck? i.e. by stopping the db

p16:
@mcollina changes done, can you have a look? 🙇

mcollina (approved):
lgtm

Issue #713, opened 2021-02-11
Jest version conflict in titus-frontend prevents build without skipping preflight check


Peterabsolon:
The frontend package cannot be built without `SKIP_PREFLIGHT_CHECK=true`. Currently that's being set by default by `npm run create:env`, but that could _potentially_ lead to issues in the long run.

```
yarn run v1.22.10
$ react-scripts build

There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "jest": "26.6.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of jest was detected higher up in the tree:

  /Users/peterabsolon/Code/titus/node_modules/jest (version: 26.6.3)

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That will permanently disable this message but you might encounter other issues.
```

PR #716, opened 2021-02-12
fix(deps): use fixed Jest version that doesn't conflict with CRA

Peterabsolon:
Closes https://github.com/nearform/titus/issues/713

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "^26.6.0",
+    "jest": "26.6.0",
     "lerna": "^3.22.1"
   },
   "husky": {
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "^26.6.3",
+    "jest": "26.6.0",
     "lerna": "^3.22.1"
   },
   "husky": {
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "26.6.0",
+    "jest": "^26.6.0",
     "lerna": "^3.22.1"
   },
   "husky": {
--- a/packages/titus-infra-aws-mira/package-lock.json
+++ b/packages/titus-infra-aws-mira/package-lock.json
@@ -8242,7 +8242,8 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/growly/-/growly-1.3.0.tgz",
       "integrity": "sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE=",
-      "dev": true
+      "dev": true,
+      "optional": true
     },
     "har-schema": {
       "version": "2.0.0",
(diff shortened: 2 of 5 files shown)


packages/titus-infra-aws-mira/package-lock.json
```diff
@@ -8242,7 +8242,8 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/growly/-/growly-1.3.0.tgz",
       "integrity": "sha1-8QdIy+dq+WS3yWyTxrzCivEgwIE=",
-      "dev": true
+      "dev": true,
+      "optional": true
```
Peterabsolon, line 8246: This is getting automatically added when I run either `npm i` or `npm ci`. I'm on npm@6. With npm@7 all package-locks are heavily reformatted/changed.

package.json
```diff
@@ -31,7 +31,7 @@
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "^26.6.3",
+    "jest": "26.6.0",
```
simoneb, line 34: can we achieve the same without removing the semver specifier from the version number? my concern is that pinning the version will prevent future upgrades, which is what we want to prevent right now but we also want to be able to automatically do them at some point.

Peterabsolon: So it turns that yes, we can have it there.

For whatever reason I was getting the same error when I left it there for the first time. It works now on a fresh installation. PR updated.

simoneb (approved):

packages/titus-infra-aws-mira/package-lock.json
simoneb: not a problem

package.json
```diff
@@ -31,7 +31,7 @@
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "^26.6.3",
+    "jest": "^26.6.0",
```
mcollina, line 34: I don't understand this change. If we need to use a specific version, remove `^`.

mcollina (changes requested):

package.json
simoneb: we need a specific version that matches the one supported by react-scripts, but we don't want to prevent them from being upgraded by renovate, we want to be able to keep bumping both, but they need to be in sync

mcollina: renovate will try to bump them anyway. This needs to be fixed because in `npm update` will update this.

simoneb: nobody is supposed to run npm update :) the other reason I would like to keep the semver specifier is to avoid making an exception for this package alone. because having it or not doesn't make a difference for any practical purposes (we have a lock file too), I like the principle of least surprise

mcollina: My point is that zapping package-lock.json and running a fresh npm install should produce a working repository. If that's not the case, then let's fix the dependency.

simoneb: I don't see a reason why the repository should work when removing a file that's essential for the correct working of the repository (package-lock.json), but if you think that needs to be supported then yes, this dependency should be pinned. I'm fairly sure many repositories with a package-lock will stop working properly if you remove the lock file, because they'll get bumps that are not compatible. But ok, not a strong opinion about this, it will work either way

mcollina: There are plenty of times where package-lock.json becomes _unmaintainable_ and a user needs to remove it. I prefer not having tech debt going to into fresh project.

Peterabsolon: Updated the PR to use the pinned version then.

mcollina (approved):
lgtm

PR #722, opened 2021-02-15
Fix Jest Dependency Tree

penx:
This PR ensures that `jest` is not hoisted to the root of the repository, so that `react-scripts` does not complain.

Fixes https://github.com/nearform/titus/pull/717#issuecomment-778468797

Jest is used here:

https://github.com/nearform/titus/blob/cc215564d5b3013e958f6c380ff52e850aa2c6e5/lib/tools.test.js#L11

Which means it had been added as a root dependency.

Sub packages that use Jest, other than `titus-frontend`, define it as a devDependency:

https://github.com/nearform/titus/blob/98bcb57473d814a2fcf128bb54a33b2b08bd5e96/packages/titus-backend/package.json#L60

https://github.com/nearform/titus/blob/ef7604083ec083a1ac959217ebfec8859f0ce3f3/packages/titus-db-manager/package.json#L48

However, in `titus-frontend`, react-scripts' default behaviour is to complain if `jest` is added as a dependency:

```
The react-scripts package provided by Create React App requires a dependency:

  "jest": "26.6.0"

Don't try to install it manually: your package manager does it automatically.
```


package.json
```diff
@@ -22,22 +22,20 @@
...
   "devDependencies": {
     "docsify-cli": "^4.4.2",
     "husky": "^4.3.8",
-    "jest": "26.6.3",
     "lerna": "^3.22.1"
   },
   "husky": {
     "hooks": {
       "pre-commit": "npm run lint:staged && npm run test:related",
-      "pre-push": "npm run lint:all && npm run test"
```
penx, line 40: `npm run test` no longer exists. It was previously only running the tests in `lib/tools.test.js`. Requiring this before pushing doesn't make much sense to me. Alternatively we could run `npm run test:all`, but this would mean a user has to have an active database server (and more?) in order to successfully push. Combined with the time taken to run all tests, this seems like an unnecessary overhead to run on `pre-push`.

lerna.json
```diff
@@ -1,7 +1,9 @@
 {
   "lerna": "2.11.0",
   "packages": [
-    "packages/*"
+    "packages/*",
+    "lib"
   ],
-  "version": "1.0.0-alpha.1"
+  "version": "1.0.0-alpha.1",
+  "nohoist": "jest"
```
penx, line 8: I need to double check this

penx, line 8: https://github.com/lerna/lerna/blob/main/commands/bootstrap/README.md#--nohoist-glob

simoneb:
I have a feeling this might not be needed in the first place. The reason is that I recently set up monorepos where one of the packages was a CRA e this wasn't needed. Is there a chance that jest doesn't need to be installed explicitly in the first place? CRA may be installing the right version already without needing to reference it in the package.json dependencies, which may solve this issue altogether.

I believe this is the package.json that's created by recent versions of CRA https://github.com/facebook/create-react-app/blob/master/packages/cra-template/template.json.

For an example of a monorepo where one of the packages is CRA and nothing fancy is needed to make it run properly see https://github.com/nearform/brokeneck/

penx:
> For an example of a monorepo where one of the packages is CRA and nothing fancy is needed to make it run properly see https://github.com/nearform/brokeneck/

In brokeneck, jest is only listed as a dependency in one project: brokeneck-react. Therefore learna doesn't hoist Jest to the repository root.

`react-scripts` is a dependency of a different project - brokeneck-html, so `react-scripts` doesn't complain about finding Jest in its path.

The issue arrises when lerna sees that jest is a shared dependency, hoists it, but it is not the exact version that react-scripts is expecting. This can be managed by either:

- ensuring your whole monorepo is in sync with the version of jest that react-scripts expects
- (I think adding Jest as a dependency alongside react-scripts with a strict version match)
- telling Lerna (or Yarn workspaces if using) not to hoist these dependencies

> CRA may be installing the right version already without needing to reference it in the package.json dependencies

There are 2 issues here:

1. In a monorepo, at least from what I have read, it is good practice to have each package self contained where possible
  - e.g.  not be dependent on dependencies that may have been hoisted from elsewhere
  - e.g. a user should ideally be able to check out the project, cd into the package that interests them, `npm i` and run that package standalone (where feasible)
2. if only one package is using react-scripts, and jest isn't mentioned elsewhere, then jest won't be hoisted to the project route

package.json
simoneb: I agree with the change you did here

simoneb (commented):
This LGTM. One question, which is the main point. We want the build to fail if you upgrade the Jest version in the frontend package, which doesn't look like it's happening here.

penx:
> We want the build to fail if you upgrade the Jest version in the frontend package

the frontend package has no dependency on jest now, and jest is not hoisted to the root, so my understanding is the only way jest could be upgraded on the frontend package now is by updating react-scripts

penx:
I pushed some `package-lock` updates but for some reason they're showing 

```diff
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```

I need to look in to this

simoneb:
> > We want the build to fail if you upgrade the Jest version in the frontend package
> 
> the frontend package has no direct dependency on jest now, and jest is not hoisted to the root, so my understanding is the only way jest could be upgraded on the frontend package now is by updating react-scripts

ah yes that should do it, good point

simoneb:
> I pushed some `package-lock` updates but for some reason they're showing
> 
> ```diff
> -  "lockfileVersion": 1,
> +  "lockfileVersion": 2,
> ```
> 
> I need to look in to this

let's keep version 1. use nvm to make sure you're running a npm version compatible with the node version we're using here, which is 14.

simoneb (approved):
All good, let's just make sure package lock is version 1

PR #729, opened 2021-02-18
chore: update mira and cdk to latest versions

nigelhanlon:
This pull request updates the Mira package to the latest available version (1.5.0-beta) with the matching AWS CDK (v1.89.0). 

A small change was needed to the way database props were passed to the Database instance.


simoneb (approved):
Nice!

PR #757, opened 2021-03-08
Fixes #756 : CSP conflict between fastify-swagger & fastify-helmet

### LINKED ISSUE #756: Swagger not working out of the box due to Content Security Policy conflict
Swagger UI includes inline JavaScript, CSS styles and images. Due to the CSP set by `fastify-helmet`, the browser refuses to evaluate these resources, and the screen remains blank.

The fix I implemented locally involves updating `fastify-swagger` and `fastify-helmet` to the latest versions and updating two files (`lib/server.js` and `index.js`) and one test file (`lib/server.test.js`) in the back end package:

1. Move `fastify-swagger` registration from `lib/server.js` to `index.js` before the `fastify-helmet` registration.
2. Use the `fastify-helmet` configuration suggested in the [`fastify-swagger` docs](https://github.com/fastify/fastify-swagger#integration).
3. Update line 8 in `lib/server.test.js` to : `expect(server.register).toHaveBeenCalledTimes(4)`.

If I find some time, I'll prepare a PR for this.

kurtmilam:
This PR fixes #756

It also upgrades all npm dependencies across the monorepo.

I can remove dependency updates for packages other than `titus-backend` if desired.


simoneb (changes requested):
Unfortunately we can't upgrade to the new lock file version because it's not compatible with the npm@6 version. You generated the new lockfiles with npm@7 instead. Because we support Node LTS we should use the version of npm that ships with that, which is npm@6.

kurtmilam:
Ah yes, that is unfortunate. It would probably be a good idea to add a note about that restriction to the contributing document, as well. I would also suggest adding a note about needing to run `npm run create:env` and any other setup that's necessary to get tests passing.

simoneb:
@kurtmilam a note about which version of npm to use should not be required because that's the safest assumption. Node developers often use nvm to handle locally installed Node versions, and lacking any specific information the only sensible default is to use the LTS node version, which we're using here. 

As for the setup guide, it's in https://nf-titus.netlify.app/#/quick-start/

packages/titus-backend/index.js
```diff
@@ -16,7 +16,18 @@ process.on('unhandledRejection', err => {
...
+    contentSecurityPolicy: {
+      directives: {
+        ...helmet.contentSecurityPolicy.getDefaultDirectives(),
+        'form-action': [`'self'`],
+        'img-src': [`'self'`, 'data:', 'validator.swagger.io'],
+        'script-src': [`'self'`].concat(instance.swaggerCSP.script),
+        'style-src': [`'self'`, 'https:'].concat(instance.swaggerCSP.style)
+      }
+    }
+  }))
```
simoneb, line 30: I wonder if, instead of moving swagger here, we shouldn't move the two registrations to lib/server.js instead

simoneb (commented):
Do you think there's a chance this change can be tested so we know we're not breaking swagger in the future?

kurtmilam:
> @kurtmilam a note about which version of npm to use should not be required because that's the safest assumption. Node developers often use nvm to handle locally installed Node versions, and lacking any specific information the only sensible default is to use the LTS node version, which we're using here.
> 
> As for the setup guide, it's in https://nf-titus.netlify.app/#/quick-start/

I use nvm to handle locally-installed Node versions.
image

v14, which I used here, is Active LTS, and is specified in the `.nvmrc` at the root of the project. Perhaps I've misunderstood your comment.

simoneb:
> > @kurtmilam a note about which version of npm to use should not be required because that's the safest assumption. Node developers often use nvm to handle locally installed Node versions, and lacking any specific information the only sensible default is to use the LTS node version, which we're using here.
> > As for the setup guide, it's in https://nf-titus.netlify.app/#/quick-start/
> 
> I use nvm to handle locally-installed Node versions.
> image
> 
> v14, which I used here, is Active LTS, and is specified in the `.nvmrc` at the root of the project. Perhaps I've misunderstood your comment.

That's correct but you must have updated npm manually, because that version of Node ships with npm@6, therefore that's the one that should be used. See https://nodejs.org/en/download/releases/

kurtmilam:
> Do you think there's a chance this change can be tested so we know we're not breaking swagger in the future?

Possibly. I haven't looked at the underlying code, but my understanding was that `fastify-swagger` is now generating a hash of the inlined JavaScript and adding the hash to the list of allowed script sources. It is my assumption that this hash is generated on demand, and will change to match whatever inline JavaScript is produced for the Swagger UI page.

I've also not thought about how to test whether this specific configuration is working. To me, that feels like testing external libraries, which I usually avoid.

kurtmilam:
> > > @kurtmilam a note about which version of npm to use should not be required because that's the safest assumption. Node developers often use nvm to handle locally installed Node versions, and lacking any specific information the only sensible default is to use the LTS node version, which we're using here.
> > > As for the setup guide, it's in https://nf-titus.netlify.app/#/quick-start/
> > 
> > 
> > I use nvm to handle locally-installed Node versions.
> > image
> > v14, which I used here, is Active LTS, and is specified in the `.nvmrc` at the root of the project. Perhaps I've misunderstood your comment.
> 
> That's correct but you must have updated npm manually, because that version of Node ships with npm@6, therefore that's the one that should be used. See https://nodejs.org/en/download/releases/

Indeed - You know `npm`, always nagging you to update, and I usually do out of habit. I'm sure I'm not the only dev who does :)

packages/titus-backend/index.js
kurtmilam: I considered doing this. It wasn't clear to me whether registering `helmet` as soon as possible was best practice. Someone chose to register it in `index.js`, rather than in `server.js`, so I figured I would defer to their decision to do so. I don't mind moving it, either. I'm simply not familiar enough with the internals of `fastify`, `helmet` or `fastify-helmet` to know whether there's a good reason to register `fastify-helmet` as early as possible.

simoneb:
> Indeed - You know `npm`, always nagging you to update, and I usually do out of habit. I'm sure I'm not the only dev who does :)

Understood. Unfortunately npm@7 introduces breaking changes so we can't use it unless everybody uses it on a specific project. It's better to not do the upgrade and wait until that version of npm is the one that's released with the version of Node that's being used, otherwise it's a hassle for everybody.

packages/titus-backend/index.js
simoneb: Knowing the codebase I don't see a good reason why it should be done here instead of lib/server.js, I'd move it there. If you then revert the package-lock file changes, this PR is good to go.

kurtmilam:
Thinking about tests - there is a requirement for `fastify-swagger` to be registered before `fastify-helmet`, unfortunately. This may be worth adding a test for.

simoneb:
> Thinking about tests - there is a requirement for `fastify-swagger` to be registered before `fastify-helmet`, unfortunately. This may be worth adding a test for.

I agree!

kurtmilam:
I'm honestly not sure how to write that test. We have `toHaveBeenCalledWith` and `toHaveBeenCalledBefore`, but we don't have a `toHaveBeenCalledWithBefore`. If someone decides to move helmet registration before swagger registration, the server won't start, so perhaps a comment and a slight change to the code to make it clear that `swagger` is required before `helmet` is enough here?

simoneb:
> I'm honestly not sure how to write that test. We have `toHaveBeenCalledWith` and `toHaveBeenCalledBefore`, but we don't have a `toHaveBeenCalledWithBefore`. If someone decides to move helmet registration before swagger registration, the server won't start, so perhaps a comment and a slight change to the code to make it more clear that `swagger` is required before `helmet` is enough here?
> 
> I went ahead and pushed the suggested change.
> 
> It would be handy if `fastify-swagger` exposed the necessary hashes in some form other than just decorating the server instance with them. My preference would be to import the hashes and add them to the helm registration callback so it wouldn't matter in which order the plugins were registered, but that's out of scope for discussion in this repository, I believe.

yeah that's good enough if this isn't easy to test

kurtmilam:
Great - I'll just round this off with the comment that I'm not super familiar with the [Encapsulation context](https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md) feature of fastify, and it's not clear to me how moving these `register` calls from `index.js` into `lib/server.js` affects their availability across different contexts.

simoneb:
> Great - I'll just round this off with the comment that I'm not super familiar with the [Encapsulation context](https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md) feature of fastify, and it's not clear to me how moving these `register` calls from `index.js` into `lib/server.js` affects the availability of the registered plugins across different contexts.
> 
> Is it possible that it's best to register a security-related plugin like `fastify-helmet` at the root context created in `index.js`, rather than in the shared context that's created in `lib/server.js`?

No, there isn't a good reason why fastify-helmet should be registered elsewhere as far as I can tell. I would move it to lib/server.js and see if anything breaks

simoneb (approved):
LGTM

simoneb (approved):
LGTM

Issue #636, opened 2020-12-01
Failing test from Mira package


feugy:
### Reproduction steps and actual behavior

On a fresh clone, run all the tests:
```shell
git clone git@github.com:nearform/titus.git
cd titus
npm i
npm run create:env
npm run db:up
npm run test:all
```

Then you'll get errors from Mira package:
```shell
titus-infra-aws-mira: FAIL cdk/core/index.test.ts
titus-infra-aws-mira:   ● Test suite failed to run
titus-infra-aws-mira:     cdk/core/database.ts:51:7 - error TS2345: Argument of type '{ engine: IInstanceEngine; instanceType: InstanceType; vpc: IVpc; vpcPlacement: { subnetType: SubnetType.PUBLIC; }; storageEncrypted: false; ... 9 more ...; securityGroups: ISecurityGroup[]; }' is not assignable to parameter of type 'DatabaseInstanceProps'.
titus-infra-aws-mira:       Object literal may only specify known properties, and 'masterUsername' does not exist in type 'DatabaseInstanceProps'.
titus-infra-aws-mira:     51       masterUsername,
titus-infra-aws-mira:              ~~~~~~~~~~~~~~
titus-infra-aws-mira: FAIL cdk/api-gw-nlb/index.test.ts
```

(repeated several times)

### Expected behavior?

All tests should pass.

### Configuration

Running on Ubuntu 20.04.1, 64x, node 14.12

simoneb:
There is a chance that it may be related to upgrading package versions automatically. I have tried to be careful to avoid bumping packages which I knew would be causing issues, like the AWS CDK ones (you can see that reflected in the Renovate configuration), but I believe mira was upgraded by a minor version, for instance.

irelandm:
Yes, generally things break if there is *any* discrepancy with CDK versions

simoneb:
@irelandm there shouldn't be any discrepancies with cdk versions:

- the version of mira that's being used is [1.4.1](https://github.com/nearform/titus/blob/master/packages/titus-infra-aws-mira/package.json#L55)
- which [uses cdk 1.67.0](https://github.com/nearform/mira/blob/v1.4.1/package.json#L27)
- and titus [uses that too](https://github.com/nearform/titus/blob/master/packages/titus-infra-aws-mira/package.json#L32)
- and I made sure that Renovate [does not try to bump cdk](https://github.com/nearform/titus/blob/master/renovate.json#L15)

[...]

PR #760, opened 2021-03-08
Rewriting getting started commands for testing

inigo-garcia:
This PR intends to remark the importance of building the frontend before running the tests. 

Also it decorates the build commands for better readability .

It may be related to #636


simoneb (commented):
Why don't we do a whole build rather than having to cd into a package and then out? There's a build:all script in the root dir

inigo-garcia:
Changed to `npm run build:all` and resolved an ambiguous test message due missing package name.

simoneb (approved):

PR #795, opened 2021-03-30
GCP deploy fixes

### LINKED ISSUE #787: Deploy on GCP issues
Followed [the steps](https://nf-titus.netlify.app/#/devops/gcp/?id=deploy-titus-on-gcp) in the documentation to deploy Titus to GCP, but some parts are still missing to get it right the first time.

- There needs to be a bucket available for TF to store state, but there is no mention of it in the docs that this needs to be created manually before running `terraform apply`. When just running this command, there is a misleading error about invalid permissions. Also, a unique bucket name needs to be provided in `main.tf` first.
- On first terraform deployment, there is an error about the enabling of the "Cloud Resource Manager API". This can be avoided by adding `gcloud services enable cloudresourcemanager.googleapis.com` to the list of commands to run to enable needed API's.
- Docs should mention the **_only_** the first part of `config.auto.tfvars` should be changed, and not the other vars. I've also changed `artifact_registry_repository_name` for example, but that is causing errors in the GH deployment pipelines. The vars. are not supported everywhere yet.
- After succesful deployment, the backend service does not start because the `AUTH0_DOMAIN` env var is missing. When setting this to `AUTH0_DOMAIN=dummy` resolves the issue, but something should be added to TF/docs for this.
- All the docs mention the project path `infra/terraform/gcp/`, but this should be `infra/gcp/terraform/`.

ovhemert:
Fixes #787 to be able to deploy on GCP without any manual changes.

### DIFF
--- a/docs/devops/gcp/README.md
+++ b/docs/devops/gcp/README.md
@@ -42,9 +42,14 @@ gcloud services enable run.googleapis.com
 gcloud services enable iam.googleapis.com
 gcloud services enable secretmanager.googleapis.com
 gcloud services enable artifactregistry.googleapis.com
+gcloud services enable cloudresourcemanager.googleapis.com
 ```
 
 ## Provision infrastruture
+
+To allow Terraform to provision services on GCP we need to give it a bucket to store it's state.
+
+In GCP web console navigate to `Cloud Storage` and create a new bucket with the `Create Bucket` option. Set a unique name for the bucket and adjust `infra/gcp/terraform/main.tf`.
 
 From `infra/gcp/terraform` folder run these commands from your terminal:
 ```sh
--- a/docs/devops/gcp/README.md
+++ b/docs/devops/gcp/README.md
@@ -47,7 +47,7 @@ gcloud services enable cloudresourcemanager.googleapis.com
 
 ## Provision infrastruture
 
-To allow Terraform to provision services on GCP we need to give it a bucket to store it's state.
+To allow Terraform to provision services on GCP we need to give it a bucket to store its state.
 
 In GCP web console navigate to `Cloud Storage` and create a new bucket with the `Create Bucket` option. Set a unique name for the bucket and adjust `infra/gcp/terraform/main.tf`.
 
--- a/docs/devops/gcp/README.md
+++ b/docs/devops/gcp/README.md
@@ -18,12 +18,12 @@ The stack serves only as an example and is expected to be adjusted or changed ba
 
 To allow Terraform to provision services on GCP we need to give it a service account with right permissions.
 
-In GCP web console navigate to `IAM & Admin / Service Accounts` and create a new service account with `Project Owner` role. Then create a key from it and download it as `key.json` into `infra/terraform/gcp`.
+In GCP web console navigate to `IAM & Admin / Service Accounts` and create a new service account with `Project Owner` role. Then create a key from it and download it as `key.json` into `infra/gcp/terraform`.
 
 
 ## Configure GCP project variables
 
-Create a `infra/terraform/gcp/config.auto.tfvars` file and fill it with adjusted content of `infra/terraform/gcp/config.auto.tfvars.sample`. You must set these variables to match your GCP project:
+Create a `infra/gcp/terraform/config.auto.tfvars` file and fill it with adjusted content of `infra/gcp/terraform/config.auto.tfvars.sample`. You must set these variables to match your GCP project:
 - `gcp_project_id`
 - `region`
 - `zone`
@@ -46,7 +46,7 @@ gcloud services enable artifactregistry.googleapis.com
 
 ## Provision infrastruture
 
-From `infra/terraform/gcp` folder run these commands from your terminal:
+From `infra/gcp/terraform` folder run these commands from your terminal:
 ```sh
 terraform init
 terraform apply
--- a/infra/gcp/terraform/backend.tf
+++ b/infra/gcp/terraform/backend.tf
@@ -3,6 +3,14 @@ resource "google_service_account" "backend" {
   display_name = "backend"
 }
 
+resource "google_project_iam_binding" "backend_iam" {
+  depends_on = [google_service_account.backend]
+  role    = "roles/cloudsql.client"
+  members = [
+    "serviceAccount:${google_service_account.backend.email}"
+  ]
+}
+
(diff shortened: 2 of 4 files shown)


docs/devops/gcp/README.md
```diff
@@ -42,11 +42,16 @@ gcloud services enable run.googleapis.com
 gcloud services enable iam.googleapis.com
 gcloud services enable secretmanager.googleapis.com
 gcloud services enable artifactregistry.googleapis.com
+gcloud services enable cloudresourcemanager.googleapis.com
 ```
 
 ## Provision infrastruture
 
-From `infra/terraform/gcp` folder run these commands from your terminal:
+To allow Terraform to provision services on GCP we need to give it a bucket to store it's state.
```
simoneb, line 50: apologies for nitpicking on this stuff. it's -> its, right?

docs/devops/gcp/README.md
```diff
@@ -42,11 +42,16 @@ gcloud services enable run.googleapis.com
...
+
+In GCP web console navigate to `Cloud Storage` and create a new bucket with the `Create Bucket` option. Set a unique name for the bucket and adjust `infra/gcp/terraform/main.tf`.
```
simoneb, line 52: I assume this is not something that can be automated, right? If unsure, you may want to check with @dgonzalez

infra/gcp/terraform/backend.tf
```diff
@@ -3,6 +3,14 @@ resource "google_service_account" "backend" {
...
 }
 
+resource "google_project_iam_binding" "backend_iam" {
+  depends_on = [google_service_account.backend]
+  role    = "roles/cloudsql.client"
+  members = [
+    "serviceAccount:${google_service_account.backend.email}"
+  ]
+}
+
```
simoneb, line 13: ah nice! to make it work I also had to give permissions to read from the secret storage, are you sure this works without that permission?

infra/gcp/terraform/backend.tf
```diff
@@ -72,6 +80,10 @@ resource "google_cloud_run_service" "backend" {
           name  = "NODE_ENV"
           value = "production"
         }
+        env {
+          name  = "AUTH0_DOMAIN"
+          value = "dummy"
+        }
```
simoneb, line 86: I would prefer if we fixed this in the source code rather than the infrastructure. IoW, if we're not using Auth0, make it so that the server doesn't require this variable to be set.

infra/gcp/terraform/db-manager.tf
```diff
@@ -64,6 +72,10 @@ resource "google_cloud_run_service" "db_manager" {
           name  = "NODE_ENV"
           value = "production"
         }
+        env {
+          name  = "AUTH0_DOMAIN"
+          value = "dummy"
+        }
```
simoneb, line 78: same comments as in the backend tf file

docs/devops/gcp/README.md
ovhemert: nitpicking is always fine

infra/gcp/terraform/backend.tf
ovhemert: It works without that extra permission. The difference I think is in the new account you needed to create, while this is adjusting existing accounts with the TF config. Probably the CloudRun service accounts already have the requires roles to read from secret store.

simoneb: Ah could be, I found find it weird though because I'm not sure why it would have implicitly one permission but not the other, but I guess this could very well be the case. Just to confirm, the secret would only be used when actually connecting to the DB, possibly just from the backend (and not db-manager). So, you're saying that this works without this permission when connecting to the DB from the backend. Do we have an API which hits the DB to confirm that it works? Because the fact we didn't have one in Titus by default was the reason why we didn't notice that the db connection wouldn't work.

ovhemert: There is an `/healthcheck` route in the backend that responds with a json. This route actually makes a call to the database to check connectivity. The resulting `db` property can be either `ok` or `fail`.
For the `db-manager`, you can POST `/db/migrate` which will either return 200 or else fail.

Both services are configured fine now after added TF config.

docs/devops/gcp/README.md
ovhemert: The state bucket is one the things that need to be pre-configured manually before TF can run.

simoneb (commented):
Overall LGTM, it's still missing the Auth0 change so that the env var doesn't need to be set unless we're using Auth0, which we're not using by default.

ovhemert:
@simoneb The backend auth providers available are: `auth0`, `azure-ad` and `cognito`. The one that is provided in `AUTH_PROVIDER` env-var is initialized. If none is specified, the fallback one is `auth0`.
This causes problems on the production deployment, since no auth provider is specified there. So, the default will be `auth0`. But to use it, it needs additional settings (like `AUTH0_DOMAIN`).
On development, `cognito` is specified in the env-vars. This does not need additional values, so it initializes fine.

We can either set the default to `cognito` for both development and production. Or we can implement a dummy (in-memory) authorization provider as the default.

simoneb:
> @simoneb The backend auth providers available are: `auth0`, `azure-ad` and `cognito`. The one that is provided in `AUTH_PROVIDER` env-var is initialized. If none is specified, the fallback one is `auth0`.
> This causes problems on the production deployment, since no auth provider is specified there. So, the default will be `auth0`. But to use it, it needs additional settings (like `AUTH0_DOMAIN`).
> On development, `cognito` is specified in the env-vars. This does not need additional values, so it initializes fine.
> 
> We can either set the default to `cognito` for both development and production. Or we can implement a dummy (in-memory) authorization provider as the default.

Understood. In that case it's ok to then initialize the env var.

infra/gcp/terraform/backend.tf
ovhemert: See [comment below](https://github.com/nearform/titus/pull/795#issuecomment-811709476), need to leave this in place.

infra/gcp/terraform/db-manager.tf
ovhemert: See [comment below](https://github.com/nearform/titus/pull/795#issuecomment-811709476), need to leave this in place.

simoneb (approved):

mcollina (approved):
lgtm

PR #855, opened 2021-05-05
Prefer package.location to package.name in (create/clean)Env fns

### LINKED ISSUE #851: createEnv / cleanEnv scripts introduce hard dependency on matching package names with containing directory names
### Background

It is my understanding that neither lerna nor yarn workspaces require a package's name and the name of the directory that contains the package to match.

In other words, it should be possible to have a package at `packages/pkg-dir/` with the package name of `"pkg-name"`.

### Problem

The `createEnv` and `cleanEnv` scripts (in `/lib/tools.js`) introduce such a requirement because they use each package's name to build the directory in which the script searches for the package's `.env.sample` file.

### Suggested Solution

`'npx lerna ls --all --json'` (used by the script) returns a JSON array of package objects. Each package object has a `name` and `location` property. Rather than using the `name` property to build each package's directory, simply use the `location` property. I have tested this solution in a live project, and it seems to work well.

### Code

Here's my local, working version of `/lib/tools.js`. This file also improves some of the logging messages and variable names while removing some complexity introduced by attempting to concatenate the package directory paths from each package's name:

```javascript
'use strict'

const fs = require('fs')
const path = require('path')
const { execSync } = require('child_process')
const chalk = require('chalk')

const warning = chalk.keyword('yellow')
const info = chalk.keyword('green')

const pkgDirs = getPkgDirs()
function getPkgDirs() {
  try {
    const packages = execSync('npx lerna ls --all --json')
    return JSON.parse(packages.toString()).map(
      ({ location }) => location
    )
  } catch (error) {
    console.log(warning('No packages found'))
    process.exit(0)
  }
}

function cleanEnv() {
  for (const pkgDir of pkgDirs) {
    const env = `${pkgDir}/.env`

    if (!fs.existsSync(env)) {
      continue
    }

    try {
      fs.unlinkSync(env)
      console.log(info(`Removed .env file at ${pkgDir}`))
    } catch (err) {
      console.log(warning(`No .env file at ${pkgDir}`))
    }
  }
}

function createEnv() {
  for (const pkgDir of pkgDirs) {
    const sample = `${pkgDir}/.env.sample`
    const env = `${pkgDir}/.env`

    if (!fs.existsSync(sample)) {
      continue
    }

    try {
      fs.accessSync(env)
      console.log(warning(`.env file at ${pkgDir} already exists`))
    } catch (err) {
      if (err && err.code === 'ENOENT') {
        fs.copyFileSync(sample, env)
        console.log(info(`Created .env file at ${pkgDir}`))
      }
    }
  }
}

module.exports = {
  cleanEnv,
  createEnv
}

```

I'll try to open a PR that fixes this issue some time this week, if someone else doesn't beat me to it.

kurtmilam:
Closes #851 

With this PR, the `create:env` and `clean:env` scripts will use each package's `location` property to look for or create corresponding `.env` / `.env.sample` files. This change makes it possible to have a package name that does not exactly equal the name of the package's containing directory.

Other minor updates include:
1. Improve logging messages.
2. Simplify code since we now have a path, rather than a name that needs to be concatenated into a path.
3. Minor 'improvements' based on developer preference.

Test Results:
image

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -25,7 +25,9 @@ function cleanEnv() {
   for (const pkgDir of pkgDirs) {
     const env = `${pkgDir}/.env`
 
-    if (!fs.existsSync(env)) continue
+    if (!fs.existsSync(env)) {
+      continue
+    }
 
     try {
       fs.unlinkSync(env)
@@ -40,7 +42,9 @@ function createEnv() {
   for (const pkgDir of pkgDirs) {
     const sample = `${pkgDir}/.env.sample`
 
-    if (!fs.existsSync(sample)) continue
+    if (!fs.existsSync(sample)) {
+      continue
+    }
 
     const env = `${pkgDir}/.env`
 
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -8,51 +8,49 @@ const chalk = require('chalk')
 const warning = chalk.keyword('yellow')
 const info = chalk.keyword('green')
 
-let packages
-
-try {
-  const output = execSync('npx lerna ls --all --json')
-  packages = JSON.parse(output.toString()).map(
-    packageItem => `../packages/${packageItem.name}`
-  )
-} catch (error) {
-  console.log(warning('No packages found'))
-  process.exit(0)
+const pkgDirs = getPkgDirs()
+function getPkgDirs() {
+  try {
+    const packages = execSync('npx lerna ls --all --json')
+    return JSON.parse(packages.toString()).map(
+      ({ location }) => location
+    )
+  } catch (error) {
+    console.log(warning('No packages found'))
+    process.exit(0)
+  }
 }
 
 function cleanEnv() {
-  for (const pathDir of packages) {
-    const env = path.resolve(__dirname, `${pathDir}/.env`)
+  for (const pkgDir of pkgDirs) {
+    const env = `${pkgDir}/.env`
 
-    if (!fs.existsSync(env)) {
-      continue
-    }
+    if (!fs.existsSync(env)) continue
 
     try {
       fs.unlinkSync(env)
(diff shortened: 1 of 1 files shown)


lib/tools.js
```diff
@@ -8,51 +8,49 @@ const chalk = require('chalk')
...
+    console.log(warning('No packages found'))
+    process.exit(0)
+  }
 }
 
 function cleanEnv() {
-  for (const pathDir of packages) {
-    const env = path.resolve(__dirname, `${pathDir}/.env`)
+  for (const pkgDir of pkgDirs) {
+    const env = `${pkgDir}/.env`
```
simoneb, line 26: shouldn't this take into account the current directory as it did before?

lib/tools.js
```diff
@@ -8,51 +8,49 @@ const chalk = require('chalk')
...
-  for (const pathDir of packages) {
-    const sample = path.resolve(__dirname, `${pathDir}/.env.sample`)
-    const env = path.resolve(__dirname, `${pathDir}/.env`)
+  for (const pkgDir of pkgDirs) {
+    const sample = `${pkgDir}/.env.sample`
 
-    if (!fs.existsSync(sample)) {
-      continue
-    }
+    if (!fs.existsSync(sample)) continue
```
simoneb, line 43: formatting: let's keep the curly braces for consistency

lib/tools.js
kurtmilam: No, `package.location` is an absolute path from root.

simoneb (approved):

lib/tools.js
kurtmilam: done

simoneb:
Thanks!

PR #886, opened 2021-05-21
fix: azure ad

### LINKED ISSUE #884: Questions about Azure AD auth plugin (titus-backend)
I believe this plugin contains some bugs, mostly centered around incorrect usage of `axios`. It is possible other auth plugins contain the same bugs, although I have not checked.

Additionally, I wanted to ask whether the plugin creation is handled correctly, since we're getting an error in this plugin when we start the back end after updating to the latest version of `fastify`, [v3.15.1](https://github.com/fastify/fastify/releases/tag/v3.15.1).

## axios usage

There seem to be at least two issues with the axios usage in this plugin, although it is possible that I'm missing something. I will note that after 'fixing' these issues in the plugin and test file locally, I was able to get tests to pass. Tests failed after erroring out before I 'fixed' the issues mentioned below.

### Issue 1 - failure to conform to the [axios API when executing requests](https://github.com/axios/axios#example)

[Source](https://github.com/nearform/titus/blob/master/packages/titus-backend/lib/plugins/auth/azure-ad/index.js#L78) 
```javascript
  return axios.get({
    url: userURL,
    headers: { Authorization: `Bearer ${body.access_token}` },
    responseType: 'json'
  })
```

I believe this should either be:
```javascript
  return axios({
    method: 'GET',
    url: userURL,
    headers: { Authorization: `Bearer ${body.access_token}` },
    responseType: 'json'
  })
```

OR
```javascript
  return axios.get(userURL, {
    headers: { Authorization: `Bearer ${body.access_token}` },
    responseType: 'json'
  })
```

I converted all `axios.get` and `axios.post` calls to the second format shown above, and this fixed many of the errors. In retrospect, converting to the first format may have been less disruptive.

### Issue 2 - failure to consider that [server response data is contained in the `data` property of the response](https://github.com/axios/axios#response-schema)

As with the first issue, this shows up in tests and plugin code. Mock responses place server data directly at the root of the response object, rather than inside a `data` property on that object. Plugin code looks for response properties directly on the response object, rather than in the `data` property of that object:

[Source](https://github.com/nearform/titus/blob/master/packages/titus-backend/lib/plugins/auth/azure-ad/index.js#L25)
```javascript
  const body = await axios.get({ url: jwksUri })
  let findBy = 'x5t'
  if (!decoded.header.x5t) {
    findBy = 'kid'
  }
  const key = body.keys.find(k => k[findBy] === decoded.header[findBy])
```

The last line in the above snippet should be:
```javascript
  const key = body.data.keys.find(k => k[findBy] === decoded.header[findBy])
```

## Plugin Creation

I'm not sure whether this is a bug. It works with the penultimate version of fastify but breaks with the latest version. Basically, the auth plugin is a plugin within a plugin, where a single, implementation-specific authorization plugin is 'inserted' into a containing `auth` plugin. I'm not certain, but it's possible there is a different issue here that's causing this to break with the latest fastify.

Here, I'm basically trying to figure out whether this is a bug in Titus or a bug introduced by the latest version of fastify.

The error we see when using the latest fastify version is:
```
[nodemon] starting `node .`
TypeError: Cannot use 'in' operator to search for '#<Object>' in undefined
    at checkExistence (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastify/lib/decorate.js:55:19)
    at checkDependencies (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastify/lib/decorate.js:69:10)
    at decorate (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastify/lib/decorate.js:25:5)
    at Object.decorateFastify [as decorate] (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastify/lib/decorate.js:46:3)
    at azureAD (/Users/kmilam/dev/renx/provider-prototype/packages/prov-backend/lib/plugins/auth/azure-ad/index.js:116:10)
    at Plugin.exec (/Users/kmilam/dev/renx/provider-prototype/node_modules/avvio/plugin.js:132:19)
    at Boot.loadPlugin (/Users/kmilam/dev/renx/provider-prototype/node_modules/avvio/plugin.js:267:10)
    at release (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastq/queue.js:147:16)
    at Object.resume (/Users/kmilam/dev/renx/provider-prototype/node_modules/fastq/queue.js:80:7)
    at Plugin.finish (/Users/kmilam/dev/renx/provider-prototype/node_modules/avvio/plugin.js:246:10)
```

The exact change in fastify that seems to cause this error with this plugin is [here](https://github.com/fastify/fastify/pull/2992/files#diff-05ed2064853562e029f1bc2fa8ee0e61b2742c2e3c4368a76b5e94833fed376eR68). If we revert just that change in our locally installed fastify, the plugin loads up again correctly.

simoneb:
Closes #884


simoneb:
@kurtmilam this fixes the azure ad implementation by providing something which, despite not extremely well architected, at least makes the existing setup work.

You were right that there were quite a few issues, starting from the way axios was used that couldn't possibly work.

package.json
```diff
@@ -31,6 +31,6 @@
   "devDependencies": {
     "docsify-cli": "^4.4.3",
     "husky": "^6.0.0",
-    "lerna": "^4.0.0"
+    "lerna": "^3.22.1"
```
gilach, line 34: Doesn't it also require an ignore update rule for renovate bot?

simoneb: yes but I'm usually quick at ignoring PRs when they come in, so I'd like to keep this repo able to upgrade once the lerna issue is solved. the problem with the ignore rule is that we'll forget it and we'll never upgrade

gilach (approved):
LGTM

PR #988, opened 2021-07-28
Align docker image for backend with current best practices

### LINKED ISSUE #986: Align backend Dockerfiles with the current security best practices
There's been a very nice [blog post by Snyk](https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/) listing some best practices for Dockerfiles. Currently, Titus backend is not using most of them. It would be useful to have the best practices applied by default, as usually, security is an afterthought. I don't think making those changes would get in anyone's way.

Vunovati:
Adjust the Dockerfile according to tips in this article:
https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/

Resolve #986

### DIFF
--- a/packages/titus-backend/Dockerfile
+++ b/packages/titus-backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:lts-alpine@sha256:fb6cb918cc72869bd625940f42a7d8ae035c4e786d08187b94e8b91c6a534dfd as builder
+FROM node:14-alpine as builder
 RUN apk add dumb-init
 ENV NODE_ENV production
 
--- a/packages/titus-backend/Dockerfile
+++ b/packages/titus-backend/Dockerfile
@@ -1,4 +1,5 @@
 FROM node:lts-alpine@sha256:fb6cb918cc72869bd625940f42a7d8ae035c4e786d08187b94e8b91c6a534dfd as builder
+RUN apk add dumb-init
 ENV NODE_ENV production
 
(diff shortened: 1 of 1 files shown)


leorossi (approved):
LGTM

packages/titus-backend/Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM node:14-alpine as builder
+FROM node:lts-alpine@sha256:fb6cb918cc72869bd625940f42a7d8ae035c4e786d08187b94e8b91c6a534dfd as builder
```
simoneb, line 1: this is the only one part that I don't necessarily agree with. as this is an accelerator, it is supposed to be future proof. using whatever node 14 lts is the latest is a good way to ensure that you get... the latest 14 lts, which is stable enough and especially may contain security fixes which are not in the revision that you pinned. let's restore it to 14-alpine

simoneb (commented):
Overall LGTM, one comment about the image tag.

Please make sure to test it locally and, if all works fine, follow the CD deployment to GCP once merged to make sure it keeps running fine on GCP

packages/titus-backend/Dockerfile
Vunovati: Makes sense, this is an accelerator, not a final product.

simoneb (approved):
LGTM

PR #996, opened 2021-07-30
Enable running migrations from the docker container

### LINKED ISSUE #995: Make it possible to run db:migrate npm script from a titus-database-manager container
Currently, it is not possible to run the container with the params `npm run db:migrate`.

This makes it impossible to run a migration unless the container is running as a service. 
That is due to missing files in the Docker container.

This is a requirement for #994

Vunovati:
Fix #995

### DIFF
--- a/packages/titus-db-manager/.dockerignore
+++ b/packages/titus-db-manager/.dockerignore
@@ -5,6 +5,7 @@
 !package.json
 !package-lock.json
 !index.js
+!migration-start.js
 !seed/**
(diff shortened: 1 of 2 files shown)


simoneb (approved):
LGTM, note (some of) the changes you did to the other Dockerfile may apply here too. Also, for future proofness (and to avoid forgetting files) copying all files may be easier (and excluding those we don't want with dockerignore perhaps)

PR #997, opened 2021-07-30
Add helmfiles for k8s deployment

### LINKED ISSUE #994: Create deployment mechanism for Kubernetes
We could use a Helm chart for Titus, which would speed things up for us when having to deploy it to Kubernetes.

A very useful feature to have would be to have the chart invoke database schema migrations from the Titus-DB-manager (by using [postgrator](https://www.npmjs.com/package/postgrator)) on deployment.

To do that, we need to replicate a pattern where one can run a one-off process that is spun up whenever we do a new deployment. An example of that on other platforms would be Heroku's release phase: https://devcenter.heroku.com/articles/release-phase

The desired flow would be:
1. Deploy the project
2. Run the DB migrations (if any), while the containers with old code are still running
3. If the migration succeeds run the containers with new code; if not, leave the old containers as they were

### Solution proposal

The release phase process can be replicated in k8s by using an approach suggested in this blog post:
https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-7-running-database-migrations/

The author suggests using a combination of Job and init containers. He also explains the other options that were considered.

A Job is used to run a one-off command in a container on the k8s cluster. In this case, it is going to run the Titus-DB-manager image with the `npm run db:migrate` command.
The init containers will be added to the pod … and they will block the start of production code containers until the Job is run to completion successfully.

The init containers can know which job to wait for by looking for a matching release id in the job name.

Vunovati:
Resolve #994 

Make sure to have the required docker images in the right registry and correct referenced.
```bash
$ cd packages/titus-backend &&  docker build -t titus-backend:latest .
$ cd packages/titus-db-manager && docker build -t titus-db-manager:latest .
$ cd packages/titus-frontend && docker build -t titus-frontend:latest .
```
Make sure your local values.yaml is setup correctly
```bash
$ cat infra/k8s/helm/values.yaml
```
To expose the titus backend and frontend services install your preferred ingress controller and enable it in values.yaml for both services. By default both services are only accessible via proxy. 

When you build the frontend docker image without custom environment variables it will try to connect titus backend at localhost:5000.

By default this helm chart is deploying a postgresql instance into the same namespace as titus itself and creates a random db password as a kubernetes secret. If you want to use your own database config externalDatabase in values.yaml.

You can now deploy it via helm command to you local or any remote kubernetes cluster:
```bash
$ cd infra/k8s/helm
$ helm upgrade -i -n titus --create-namespace --values values.yaml --kubeconfig ${PATH_TO_YOUR_KUBECONFIG} titus .
```


infra/k8s/helm/Titus/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
simoneb, line 1: this comment is about file naming: do we need the Titus folder in the directory structure? We are in the titus repository, so this can only be about Titus. Let's remove it unless there are reasons to keep it there

david-freistrom: We do not really need it

docs/devops/_sidebar.md
```diff
@@ -7,4 +7,5 @@
         - [Deploy on GCP](/devops/gcp/)
         - [Deploy on AWS with Mira](https://github.com/nearform/titus/tree/master/packages/titus-infra-aws-mira)
         - [Deploy Titus on Azure](/devops/azure/?id=Deploy-Titus-on-Azure)
+        - [Deploy Titus on Kubernetes](/devops/k8s/?id=Deploy-Titus-on-Kubernetes)
```
simoneb, line 10: remove "on Titus" as it's kinda redundant (Leo already removed it for the Azure one in another PR)

docs/devops/azure/_sidebar.md
```diff
@@ -14,4 +14,5 @@
             - [Create the pipelines in Azure DevOps](/devops/azure/?id=create-the-pipelines-in-azure-devops)
             - [Update the Container images](/devops/azure/?id=update-the-container-images)
             - [Verifying Azure deployment](/devops/azure/?id=verifying-azure-deployment)
+        - [Deploy Titus on Kubernetes](/devops/k8s/?id=Deploy-Titus-on-Kubernetes)
```
simoneb, line 17: same as comment above

docs/devops/gcp/_sidebar.md
```diff
@@ -14,4 +14,5 @@
             - [Verifying GCP deployment](/devops/gcp/?id=verifying-gcp-deployment)
         - [Deploy on AWS with Mira](https://github.com/nearform/titus/tree/master/packages/titus-infra-aws-mira)
         - [Deploy Titus on Azure](/devops/azure/?id=Deploy-Titus-on-Azure)
+        - [Deploy Titus on Kubernetes](/devops/k8s/?id=Deploy-Titus-on-Kubernetes)
```
simoneb, line 17: same as comment above

docs/devops/k8s/_sidebar.md
```diff
@@ -0,0 +1,18 @@
+- [Home](/)
+- [Quick Start Guide](/quick-start/)
+- [Developers](/developers/)
+- [DevOps](/devops/?id=devops)
+    - [Overview](/devops/?id=overview)
+    - [Deploy Titus](/devops/?id=deploy-titus)
+        - [Deploy Titus on GCP](/devops/gcp/?id=Deploy-Titus-on-GCP)
+        - [Deploy on AWS with Mira](https://github.com/nearform/titus/tree/master/packages/titus-infra-aws-mira)
+        - [Deploy Titus on Azure](/devops/azure/?id=Deploy-Titus-on-Azure)
+        - [Deploy Titus on Kubernetes](/devops/k8s/?id=Deploy-Titus-on-Kubernetes)
```
simoneb, line 10: same as comment above

infra/k8s/helm/templates/db-migrate-job.yaml
```diff
@@ -0,0 +1,66 @@
...
+          - name: PGPASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "titus.database.secretName" . }}
+                key: postgresql-password
+      containers:
+      - name: titus-db-migrate
+        image: "titus-db-manager:latest"
+        imagePullPolicy: IfNotPresent
+        #command: ["until pg_isready -d $PG_DB -h $PG_HOST -p $PG_PORT -U $PG_USER; do echo Waiting for postgres; sleep 2; done;"]
```
simoneb, line 37: do we need this comment?

simoneb (approved):
LGTM. Let's get @leorossi to review this as well and try it out before merging.

simoneb (approved):
LGTM. Let's merge. @leorossi if you could please try this out in your own time to see how it's done and confirm it works it would be great.

Issue #1004, opened 2021-08-03
Support Azure KeyVault to store DB password in Azure deployment


simoneb:

leorossi:
#1010  has been merged, and covers `titus-backend`, next step is working on the infrastructure side, creating a new KV which will store the password

PR #1010, opened 2021-08-04
Use Azure Key Vault to store Postgres password

leorossi:
Refs: #1004

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,5 @@
     "husky": "^7.0.1",
     "lerna": "^3.22.1"
   },
-  "dependencies": {
-    "fastify-secrets-azure": "^1.0.0"
-  }
+  "dependencies": {}
 }
--- a/package.json
+++ b/package.json
@@ -32,5 +32,8 @@
     "docsify-cli": "^4.4.3",
     "husky": "^7.0.1",
     "lerna": "^3.22.1"
+  },
+  "dependencies": {
+    "fastify-secrets-azure": "^1.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -32,6 +32,5 @@
     "docsify-cli": "^4.4.3",
     "husky": "^7.0.1",
     "lerna": "^3.22.1"
-  },
-  "dependencies": {}
+  }
 }
--- a/packages/titus-backend/.env.sample
+++ b/packages/titus-backend/.env.sample
@@ -48,7 +48,8 @@ SECRETS_PG_PASS=PG_PASS
 #
 # This is an example for Azure Key Vault
 #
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -32,5 +32,8 @@
     "docsify-cli": "^4.4.3",
     "husky": "^7.0.1",
     "lerna": "^3.22.1"
+  },
+  "dependencies": {
+    "fastify-secrets-azure": "^1.0.0"
```
simoneb, line 37: wrong place where to install this dependency

packages/titus-backend/.env.sample
```diff
@@ -45,6 +45,10 @@ SECRETS_PG_PASS=PG_PASS
 #
 # SECRETS_STRATEGY=aws
 # SECRETS_PG_PASS=DB_PASSWORD
-
+#
+# This is an example for Azure Key Vault
+#
+# SECRETS_STRATEGY=azr
```
simoneb, line 51: let's name the strategy azure for consistency

simoneb (commented):
Ok the approach looks good, please fix the comments, remove logging, fix linting issues reported by github

packages/titus-backend/.env.sample
leorossi: Will do! (btw I used 3 chars for consistency 😅)

simoneb: fair enough, a different definition of consistency. let's stick to the "official" names: gcp, aws, azure

package.json
```diff
@@ -32,5 +32,6 @@
     "docsify-cli": "^4.4.3",
     "husky": "^7.0.1",
     "lerna": "^3.22.1"
-  }
+  },
+  "dependencies": {}
```
simoneb, line 36: remove!

packages/titus-db-manager/lib/secrets-manager.js
```diff
@@ -26,6 +27,15 @@ async function secretsManager(server) {
...
     )
   }
+
+  if (config.secretsManager.strategy === 'azure') {
+    // add required options
+    const [vaultName, secretName] =
+      config.secretsManager.secrets.dbPassword.split('|')
+    config.secretsManager.secrets.dbPassword = secretName
+    config.secretsManager.clientOptions = { vaultName }
+    console.log(config.secretsManager)
```
simoneb, line 37: remove

simoneb (changes requested):
see comments

simoneb (approved):
LGTM. Next step would be to change the Azure deployment configuration to use this and make sure it works.

PR #1014, opened 2021-08-05
Renames variable and update Azure documentation

leorossi:
Updates Azure documentation with some missing parts (like the providers needed to be registered in the subscription).

Updates also some other parts of the devops documentation, like the sidebar which was not consistent across multiple pages.


infra/azure/terraform/main.tf
```diff
@@ -9,8 +9,8 @@ terraform {
 
 terraform {
   backend "azurerm" {
-    resource_group_name  = "***********************"
-    storage_account_name = "YOUR STORAGE ACCOUNT"
+    resource_group_name  = "titus-rg"
+    storage_account_name = "titusonazure"
```
simoneb, line 13: should these be dynamic?

simoneb (commented):
Since you removed the input.tfvars file, do we need to update the instructions or they were already correct and the file should have not been there in the first place?

infra/azure/terraform/main.tf
leorossi: > Since you removed the input.tfvars file, do we need to update the instructions or they were already correct and the file should have not been there in the first place?

I don't think the input file should be on git. There is a `sample-input.tfvars` which can be copy-pasted and customized with the information needed

leorossi: oh sorry, variables are not allowed into the `backend` section, so they must be hardwired

simoneb: anything that may likely be customized (and I believe these two are) should be in variables or anywhere it makes sense to allow users to customize them. let's avoid hardcoded values which we know shouldn't be hardcoded

docs/devops/azure/README.md
```diff
@@ -32,14 +32,27 @@ Use your credentials to make sure you'll be able to run Terraform against your A
...
+Be sure that the following Resource Providers have been enabled in your subscription
+- `Microsoft.KeyVault`
+- `Microsoft.ContainerRegistry`
+- `Microsoft.ContainerInstance`
+
+You can enable with `az` cli command
+```
+$ az provider register --namespace 'Microsoft.KeyVault'
+$ az provider register --namespace 'Microsoft.ContainerRegistry'
+$ az provider register --namespace 'Microsoft.ContainerInstance'
```
simoneb, line 46: don't include the leading $ in the commands here and below

infra/azure/terraform/main.tf
simoneb: ok but even if they are not allowed, shall we hardcode them to values that we know are not portable?

simoneb:
@leorossi please address the comments and please include a description in the PR so we know what it contains and we can look it up in the future

infra/azure/terraform/main.tf
leorossi: Actually storing terraform state in a storage account is up to the user, so we can comment this part and use the local storage (as we do in AWS)

leorossi:
@simoneb the PR should be complete now for a review

simoneb:
> @simoneb the PR should be complete now for a review

Please use this button to notify this

image

simoneb (approved):

PR #1020, opened 2021-08-09
Updates SECRETS_PG_PASS env value

### LINKED ISSUE #1004: Support Azure KeyVault to store DB password in Azure deployment

leorossi:
Containerized apps in Azure now uses by default `azure` secrets strategy and the correct value for db password.

Closes #1004

### DIFF
--- a/infra/azure/terraform/backend.tf
+++ b/infra/azure/terraform/backend.tf
@@ -33,7 +33,6 @@ resource "azurerm_container_group" "titus-backend-containergroup" {
       "PG_PORT"     = "5432"
       "PG_DB"       = azurerm_postgresql_database.titus-db.name
       "PG_USER"     = "${var.db_user}@${azurerm_postgresql_server.titus-db-server.name}"
-      "PG_PASS"     = azurerm_key_vault_secret.titus-db-password.value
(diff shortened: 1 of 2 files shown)


infra/azure/terraform/backend.tf
```diff
@@ -34,8 +34,8 @@ resource "azurerm_container_group" "titus-backend-containergroup" {
       "PG_DB"       = azurerm_postgresql_database.titus-db.name
       "PG_USER"     = "${var.db_user}@${azurerm_postgresql_server.titus-db-server.name}"
       "PG_PASS"     = azurerm_key_vault_secret.titus-db-password.value
```
simoneb, line 36: just as a sanity check, this variable is used for what?

leorossi: removed, it is used only when `SECRETS_STRAGEGY == env`

simoneb: we also have one similarly named in db-manager, can you check if that one is used as well?

leorossi: checked and not used

simoneb: shall we remove it then?

simoneb (approved):

PR #1034, opened 2021-08-12
add .env.sample copy ebfore build

david-freistrom:
add cp .env.sample .env to docker build task


infra/gcp/terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,56 @@
+# This file is maintained automatically by "terraform init".
```
simoneb, line 1: do we need this file? if so, why was it not there before?

david-freistrom: it was not needed before. We changed it when we provide k8s support.

simoneb (approved):

PR #1089, opened 2021-08-30
add gitlab ci for building containers

ronandoolan2:
Adding gitlab ci file for building 3 different containers and pushing to ecr


leorossi:
Some env variables are needed in order to build, for instance, frontend. They are listed in the [.env.example](https://github.com/nearform/titus/blob/master/packages/titus-frontend/.env.sample) file. 

I think those vars should be added in the environment for each package

simoneb (approved):
LGTM, let's merge and then see if gitlab runs this. I created a mirror at https://gitlab.com/simone.busoli.nf/titus

PR #1141, opened 2021-09-14
feat: use close-with-grace for graceful server shutdown

### LINKED ISSUE #1137: use close-with-grace
https://github.com/mcollina/close-with-grace

simoneb:
Inspired from how it's used in https://github.com/fastify/fastify-cli/blob/master/templates/eject/server.js

Closes #1137


mcollina (approved):
lgtm

PR #1177, opened 2021-09-27
add titus-backend-typescript package

capaj:
I think this might be slightly improved by moving couple of shared files into a shared folder so that we can use it from raw JS backend and TS backend.


packages/titus-backend-typescript/lib/config/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  .prop('AD_APP_ID', S.string())
+  .prop('AD_SECRET', S.string())
+  .prop('SECRETS_STRATEGY', S.string())
+  .prop('SECRETS_PG_PASS', S.string().required())
+  .prop('HEALTHCHECK_URL', S.string().default('/healthcheck'))
+  .prop('HEALTHCHECK_MAX_HEAP_USER', S.number().default(768 * 1024 * 1024)) // arbitrary, 768 MB of RAM
+  .prop('HEALTHCHECK_MAX_RSS', S.number().default(1024 * 1024 * 1024)) // arbitrary, 1 GB of RAM
+  .prop('HEALTHCHECK_MAX_EVENT_LOOP_UTILIZATION', S.number().default(0.98))
+
+interface IEnv {
```
capaj, line 36: if we would use a tool like https://github.com/sinclairzx81/typebox for the schema definition, we wouldn't have to define the schema second time here. I have tried this, but I bumped into problems when compiling the json schema with ajv.

simoneb (approved):
LGTM. Clearly there's a lot of duplication but it's hard to get away from it.

simoneb:
@capaj let's way for a review from @jimmymintzer before merging

packages/titus-backend-typescript/lib/plugins/auth/azure-ad/azure-ad.test.js
```diff
@@ -0,0 +1,325 @@
+import fastify from 'fastify'
```
capaj, line 1: I wanted to port this file to TS as well, but decided against it in the end. Typing mocks is always very hard and for specs it doesn't really mattter-you still benefit from TS types even in JS files for IDE autocomplete and such.

packages/titus-backend-typescript/lib/plugins/auth/cognito/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import buildGetJwks from 'get-jwks'
+
+function authenticate(request) {
+  return request.jwtVerify()
+}
+
+async function cognito(server, options) {
+  const getJwks = buildGetJwks()
+
+  server.register(require('fastify-jwt'), {
```
jimmymintzer, line 11: NIT: For consistency, I think it's best to always use `import` vs mixing both `import` and `require`

jimmymintzer (commented):
Nice, excited to get this merged in! 

Most of the comments I have are around taking the typings a bit further. 

A couple other things I noticed:
1) Running `npm run start:all` at the root of the project will start both BE and since the TS one takes longer to startup will fail, as port 5000 is already running from `titus-backend`
2) Since the code ends up being compiled, `titus-backend-typescript/src` makes more sense to me than `titus-backend-typescript/lib`
3) When running the `npm run build` command in `titus-backend-typescript`, if I try to execute the compiled code, it gives an error because there's no `.env` file. If this is the expected behavior, we should add documentation around it

packages/titus-backend-typescript/lib/plugins/auth/cognito/index.ts
capaj: I totally agree. I haven't noticed this require. Will fix this

packages/titus-backend-typescript/lib/plugins/auth/azure-ad/azure-ad.test.js
capaj: yes that was the first thing I have tried. Still even when I used that, there are like 12 places where typescript is screaming with errors. I will publish it into a branch for you to take a look

packages/titus-backend-typescript/lib/plugins/auth/auth0/auth0-routes.ts
```diff
@@ -0,0 +1,91 @@
...
+          type: 'object',
+          properties: {
+            username: { type: 'string' },
+            password: { type: 'string' }
+          }
+        }
+      },
+      handler: async ({ log, body: { username, password } }, reply) => {
+        try {
+          const { data } = await axios({
```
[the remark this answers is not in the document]
capaj: Added now.

packages/titus-backend-typescript/lib/plugins/auth/cognito/index.ts
capaj: pushed now

capaj:
> I have seen an issue with the config previously where I had to wrap the object in Type.Strict to remove any extra properties typebox was adding to the schema

oh this definitely solves the problem! Thanks for this tip!

packages/titus-backend-typescript/lib/plugins/auth/auth0/auth0-routes.ts
```diff
@@ -0,0 +1,91 @@
...
+      schema: {
+        tags: ['auth0'],
+        body: {
+          type: 'object',
+          properties: {
+            username: { type: 'string' },
+            password: { type: 'string' }
+          }
+        }
+      },
```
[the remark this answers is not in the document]
capaj: added typebox schema for this

packages/titus-backend-typescript/lib/server.ts
```diff
@@ -0,0 +1,39 @@
+import path from 'path'
+
+import helmet from 'fastify-helmet'
+import swagger from 'fastify-swagger'
+import autoLoad from 'fastify-autoload'
+import fp from 'fastify-plugin'
+
+async function plugin(server, config) {
```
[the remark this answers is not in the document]
capaj: I have refactored all the plugin definitions

capaj:
> When running the npm run build command in titus-backend-typescript, if I try to execute the compiled code, it gives an error because there's no .env file. If this is the expected behavior, we should add documentation around it

I think this is already documented here:
https://nf-titus.netlify.app/#/quick-start/

> Since the code ends up being compiled, titus-backend-typescript/src makes more sense to me than titus-backend-typescript/lib

done

> Running npm run start:all at the root of the project will start both BE and since the TS one takes longer to startup will fail, as port 5000 is already running from titus-backend

I've added a note into the docs here: https://github.com/nearform/titus/pull/1177/commits/0576a0fac2850e4a687f6e4e15e132fe6c50c066

@jimmymintzer please can you do another round? Your notes should be adressed.

packages/titus-backend-typescript/src/config/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  },
+  enableAdmin: config.ENABLE_ADMIN,
+  pgPlugin: {
+    host: config.PG_HOST,
+    port: config.PG_PORT,
+    database: config.PG_DB,
+    user: config.PG_USER,
+    poolSize: 10,
+    idleTimeoutMillis: 30000
+  },
```
jimmymintzer, line 86:
```ts
import { PoolConfig } from 'pg'
import { PostgresPluginOptions } from 'fastify-postgres'
```

```suggestion
  } as PostgresPluginOptions & PoolConfig,
```

packages/titus-backend-typescript/src/config/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  underPressure: {
+    maxHeapUsedBytes: config.HEALTHCHECK_MAX_HEAP_USER,
+    maxRssBytes: config.HEALTHCHECK_MAX_RSS,
+    maxEventLoopUtilization: config.HEALTHCHECK_MAX_EVENT_LOOP_UTILIZATION,
+    exposeStatusRoute: {
+      url: config.HEALTHCHECK_URL,
+      routeResponseSchemaOpts
+    }
+  },
```
jimmymintzer, line 95:
```ts
import underPressurePlugin from 'under-pressure'
```

```suggestion
  } as underPressurePlugin.UnderPressureOptions,
```

packages/titus-backend-typescript/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  )
+
+  server.addHook('onClose', async (instance, done) => {
+    closeListeners.uninstall()
+    done()
+  })
+
+  try {
+    await server.listen(config.server as any as { port: number; host?: string })
```
jimmymintzer, line 46:
```suggestion
    await server.listen(config.server as ConfigOptions['server'])
```

packages/titus-backend-typescript/src/config/index.ts
```diff
@@ -0,0 +1,130 @@
+import path from 'path'
+
+import envSchema from 'env-schema'
+import { Static, Type } from '@sinclair/typebox'
+
+const envJsonSchema = Type.Strict(
+  Type.Object({
+    NODE_ENV: Type.String(),
+    API_HOST: Type.String(),
+    API_PORT: Type.String(),
```
jimmymintzer, line 10:
```suggestion
    API_PORT: Type.Number(),
```

jimmymintzer (commented):
Looks good! Added some comments on a few other ways where we can take the typings further. There may be a better approach for the configuration typings (i.e. `ConfigOptions`) but this is one option that seems to work well

packages/titus-backend-typescript/src/plugins/auth/auth0/auth0-routes.ts
```diff
@@ -0,0 +1,100 @@
...
+      url: '/auth',
+      schema: {
+        tags: ['auth0'],
+        security: [
+          {
+            apiKey: []
+          }
+        ]
+      },
+      // @ts-expect-error
```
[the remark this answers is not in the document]
capaj: I had error on this line before. Will remove

packages/titus-backend-typescript/package.json
```diff
@@ -0,0 +1,89 @@
+{
+  "name": "titus-backend-typescript",
+  "version": "1.0.0-alpha.1",
+  "private": true,
+  "description": "The backend for the Titus project using fastify",
+  "main": "dist/index.js",
```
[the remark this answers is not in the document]
capaj: the `js` file in src does not exist.
To me pointing main to a `ts` file does not make much sense. 
This is just an npm field useful for node.js libraries. This is not a library, so maybe we could omit the field entirely?

packages/titus-backend-typescript/src/plugins/auth/azure-ad/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  ].join('')
+
+  return axios({
+    url: userURL,
+    headers: { Authorization: `Bearer ${body.access_token}` },
+    responseType: 'json'
+  })
+}
+
+const azureAD: FastifyPluginAsync<{ auth: any }> = async (server, { auth }) => {
```
jimmymintzer, line 97:
```suggestion
const azureAD: FastifyPluginAsync<typeof configOptions> = async (server, { auth }) => {
```
where `import configOptions from '../../../config'`

packages/titus-backend-typescript/src/config/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  },
+  jwt: {
+    secret: config.JWT_SECRET
+  },
+  secretsManager: {
+    strategy: config.SECRETS_STRATEGY,
+    secrets: {
+      dbPassword: config.SECRETS_PG_PASS
+    },
+    clientOptions: null as any
```
jimmymintzer, line 130: can this stay as `null`?
```suggestion
    clientOptions: null
```

packages/titus-backend-typescript/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const server = Fastify(config.fastify)
+  server.register(startServer, config)
+
+  const closeListeners = closeWithGrace(
+    { delay: 2000 },
+    async ({
+      signal,
+      manual,
+      err
+    }:{ err?: Error, signal?: any, manual?: boolean }) => {
```
jimmymintzer, line 25:
```suggestion
    }:{ err?: Error, signal?: Signals, manual?: boolean }) => {
```
where `Signals` is 
```
import closeWithGrace, { Signals } from 'close-with-grace'
```

simoneb (approved):
LGTM. If everybody's happy, I am happy 🚀

PR #1337, opened 2021-11-16
fix: postgrator upgrade

giacomorebonato:
- changes for upgrading postgrator

Now Postgrator config requires a`execQuery` property where to pass the db client instance.

https://github.com/rickbergfalk/postgrator


simoneb (approved):
🚀

Eomm (approved):

Issue #924, opened 2021-06-16
Add schema-awareness and support for multiple migration directories to titus-db-manager


kurtmilam:
# Problem

As currently implemented, `titus-db-manager` has some serious issues relating to schemas and migration directories.

The most pressing issue is that the service's test migrations are applied to the `public` schema, the service's tests drop and recreate the `public` schema, and postgrator schema migration versions are stored in `public.schema_migrations`.

That's obviously not good for production or even local developer environments. There's no way to specify another directory of migration files than the current `migrate/migrations` directory (where the test migrations are stored), so, currently, you either have to disable tests or expect any migrations you've run and test data you've added to your database to be deleted each time you run tests.

Additionally, there is no way to specify the schema to which migrations should be applied (other than within the migrations themselves) or in which the migration versions should be stored. Even if I manually specify a different target schema in migrations, the version information will still be saved in the `public` schema, which will be dropped the next time tests are run.

Furthermore, it is common to use multiple schemas in a PostgreSQL database. A common use-case is supporting multitenancy by creating a separate schema for each tenant. Since it's not possible to tell this service dynamically which schema to apply migrations to, the multitenancy use case is not supported at all.

# Proposed Solution

1. Have the service (CLI and web API) accept two optional parameters with sane defaults:
 - `schema` (I propose we default to the user name defined in `.env`)
 - `migrationDir` (I propose we default to `schema`)
2. Move test migrations to `migrate/test/` and have test migrations and schema versions applied to the `test` schema, which can be dropped and recreated without touching application schemas.

# New Behavior

This proposed solution would separate the test migrations and schema from the working migrations and schema(s). It would also allow the user to apply a set of migrations stored in a single folder to multiple schemas (e.g. in the schema-per-tenant configuration). 

If a user runs `npm run db:migrate schemaName dirName` or supplies a schema and directory name in a `POST` to `/migrate`, the service will use the specified schema and migration directory names. (e.g. `npm run db:migrate tenant_1 main` would apply to the `tenant_1` schema the migrations in `migrate/main`.

If, however, a user runs `npm run db:migrate` or `POST`s to `/migrate` without supplying a schema or directory name, the service will default to trying to apply migrations in a directory matching the user's name to a schema matching the user's name (as defined in the `PG_USER` `.env` variable). I believe this is a good way to handle things, since the default `search_path` for a PostgreSQL connection is `"$user", public`. Another option would be to default to a folder and schema called `public` if the user doesn't supply either argument to the service.

[...]

PR #1432, opened 2021-12-10
add: configurable pg schema

Eomm:
implements #924 

TODOs:

- [x] add `schema` input to the API
- [x] document the new `schema` argument
- [x] document and test the new `dir` argument (it will not be added to the endpoint)

### DIFF
--- a/docs/developers/packages/titus-db-manager/README.md
+++ b/docs/developers/packages/titus-db-manager/README.md
@@ -69,7 +69,17 @@ Use the following commands with your Titus database:
 |`npm run db:seed` | Seed the database with dev data from `tools/migrations/seed_dev` with [postgrator].|
 |`npm run db:truncate` | Apply SQL truncate scripts with `psql` CLI against your database.|
 
+### Database `schema`
 
+Optionally, you can provide to the `db:migrate` two extra positional arguments:
+
+```bash
+npm run db:migrate [schema] [directory]
+# eg: npm run db:migrate titus-app /titus-app
+```
+
+- schema: the database's schema to migrate. Default is [`public`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC).
+- migrations' directory: where the migration scripts are located. Default `/migrations`
 
 [Jest]: https://jestjs.io
 [ESLint]: https://eslint.org
--- a/packages/titus-db-manager/index.js
+++ b/packages/titus-db-manager/index.js
@@ -37,9 +37,18 @@ async function run() {
       password
     }
 
-    await start(process.argv[2] || 'migrate', credentials)
+    const action = process.argv[2] || 'migrate'
+    // default PG schema https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC
+    const schema = process.argv[3] || 'public'
+    const customDir = process.argv[4] || '/migrations'
+
+    await start(action, credentials, {
+      logger,
+      schema,
+      dir: customDir
+    })
   } catch (err) {
-    logger.error('An Error has occurred, stopping', err)
+    logger.error(err, 'An Error has occurred, stopping')
     process.exit(1)
   }
 }
--- a/packages/titus-db-manager/index.js
+++ b/packages/titus-db-manager/index.js
@@ -38,7 +38,6 @@ async function run() {
     }
 
     const action = process.argv[2] || 'migrate'
-    // default PG schema https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC
     const schema = process.argv[3] || 'public'
     const customDir = process.argv[4] || '/migrations'
 
--- a/packages/titus-db-manager/index.js
+++ b/packages/titus-db-manager/index.js
@@ -37,14 +37,16 @@ async function run() {
       password
     }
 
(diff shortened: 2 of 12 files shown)


simoneb:
"implements" is not a recognized keyword to link issues. use one of the recognized ones to link the issue please https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

docs/developers/packages/titus-db-manager/README.md
```diff
@@ -69,7 +69,17 @@ Use the following commands with your Titus database:
 |`npm run db:seed` | Seed the database with dev data from `tools/migrations/seed_dev` with [postgrator].|
 |`npm run db:truncate` | Apply SQL truncate scripts with `psql` CLI against your database.|
 
+### Database `schema`
 
+Optionally, you can provide to the `db:migrate` two extra positional arguments:
```
simoneb, line 74:
```suggestion
Optionally, you can provide to the `db:migrate` command two extra positional arguments:
```

docs/developers/packages/titus-db-manager/README.md
```diff
@@ -69,7 +69,17 @@ Use the following commands with your Titus database:
...
+
+```bash
+npm run db:migrate [schema] [directory]
+# eg: npm run db:migrate titus-app /titus-app
+```
+
+- schema: the database's schema to migrate. Default is [`public`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC).
+- migrations' directory: where the migration scripts are located. Default `/migrations`
```
simoneb, line 82:
```suggestion
- `migrations`: directory where the migration scripts are located. Default `/migrations`
```

docs/developers/packages/titus-db-manager/README.md
```diff
@@ -69,7 +69,17 @@ Use the following commands with your Titus database:
...
+### Database `schema`
 
+Optionally, you can provide to the `db:migrate` two extra positional arguments:
+
+```bash
+npm run db:migrate [schema] [directory]
+# eg: npm run db:migrate titus-app /titus-app
+```
+
+- schema: the database's schema to migrate. Default is [`public`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC).
```
simoneb, line 81:
```suggestion
- `schema`: the database schema to migrate. Default is [`public`](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC).
```

packages/titus-db-manager/index.js
```diff
@@ -37,9 +37,17 @@ async function run() {
       password
     }
 
-    await start(process.argv[2] || 'migrate', credentials)
+    const action = process.argv[2] || 'migrate'
+    const schema = process.argv[3] || 'public'
+    const customDir = process.argv[4] || '/migrations'
```
simoneb, line 42: nit: what about a nice destructuring statement here? :)

packages/titus-db-manager/index.js
```diff
@@ -37,9 +37,17 @@ async function run() {
...
+
+    await start(action, credentials, {
+      logger,
+      schema,
+      dir: customDir
+    })
   } catch (err) {
-    logger.error('An Error has occurred, stopping', err)
+    logger.error(err, 'An Error has occurred, stopping')
```
simoneb, line 50: good catch!

packages/titus-db-manager/lib/routes.js
```diff
@@ -11,13 +11,35 @@ const Migrate = require('../migrate')
 const Truncate = require('../truncate')
 const { version } = require('../package')
 
+const inputSchema = {
```
simoneb, line 14: are we not using fluent-json-schema?

packages/titus-db-manager/lib/routes.js
```diff
@@ -11,13 +11,35 @@ const Migrate = require('../migrate')
...
+  oneOf: [
+    { type: 'null' },
+    {
+      description: 'Optional schema to process',
+      type: 'object',
+      properties: {
+        schema: {
+          type: 'string',
+          minLength: 1,
+          maxLength: 64
```
simoneb, line 24: I wouldn't really put this validation here. let's delegate it to pg, they know how long it can be

packages/titus-db-manager/lib/routes.js
```diff
@@ -11,13 +11,35 @@ const Migrate = require('../migrate')
...
+        }
+      }
+    }
+  ]
+}
+
 const config = require('./config')
 async function dbRoutes(server) {
+  server.decorateRequest('pgSchema', function pgSchema() {
+    return (this.body && this.body.schema) || 'public'
```
simoneb, line 34: why do we need this decorator? why not put the default value in the schema definition instead?

simoneb (commented):
A few comments but overall LGTM. I would nonetheless like to see some of the comments addressed, not necessarily all (because I am not 100% that all are correct).

packages/titus-db-manager/lib/routes.js
Eomm: we can't use the `default` value here because we are accepting a `null` body to support the old interface.
If we remove the `null` body, fastify triggers an error when a client does an HTTP POST without an empty JSON body `{}`.

simoneb (approved):

Issue #1453, opened 2021-12-20
Fix broken build


simoneb:
Appartently there's a localhost url in some package-lock file. Let's fix it.

PR #1454, opened 2021-12-20
Fix package refs to localhost

ovhemert:
Frontend package lock file has some references to localhost.
Restored that to point to https://registry.npmjs.org.

See issue #1453


simoneb (approved):

simoneb:
@ovhemert can you please link the issue?

Issue #1465, opened 2022-01-03
Storybook deployment fails


simoneb:
https://github.com/nearform/titus/runs/4674913701?check_suite_focus=true

PR #1466, opened 2022-01-03
Feature/optimize docker images

mablanco:
Added several optimizations to workflows and Docker images:

- Vulnerabilities scan as a reusable workflow
- Size reduction (where possible)
- Follow best practices when building Docker images

`storybook` workflow is currently failing (waiting for #1465 to be solved).

### DIFF
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -8,7 +8,6 @@ on:
     paths:
       - 'packages/titus-backend/**'
       - '.github/workflows/backend-cd-gcp.yml'
-      - '.github/workflows/grype.yml'
 
 env:
   DOCKER_SAVE_ARTIFACT: docker-save.tar
@@ -82,10 +81,8 @@ jobs:
           name: ${{ env.DOCKER_SAVE_ARTIFACT }}
           path: /tmp
 
-      - name: Load frontend image
-        run: |
-          docker load --input /tmp/$DOCKER_SAVE_ARTIFACT
-          docker image ls
+      - name: Load Docker image
+        run: docker load --input /tmp/$DOCKER_SAVE_ARTIFACT
 
       # - name: Push container
       #   working-directory: ./packages/titus-backend
@@ -103,4 +100,4 @@ jobs:
       - name: Delete Docker artifacts
         uses: geekyeggo/delete-artifact@v1
         with:
-          name: ${{ env.DOCKER_SAVE_ARTIFACT }}   
+          name: ${{ env.DOCKER_SAVE_ARTIFACT }}
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -1,6 +1,7 @@
 name: Backend CD to GCP
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -63,7 +63,7 @@ jobs:
 
   docker-scan:
     needs: build-titus-backend
-    uses: nearform/titus/.github/workflows/grype.yml@feature/optimize-docker-images
+    uses: nearform/titus/.github/workflows/grype.yml@master
     with:
       docker-image: docker-save.tar
       scan-report: titus-backend
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - feature/optimize-docker-images
     paths:
       - 'packages/titus-backend/**'
       - '.github/workflows/backend-cd-gcp.yml'
@@ -84,18 +83,18 @@ jobs:
       - name: Load Docker image
         run: docker load --input /tmp/$DOCKER_SAVE_ARTIFACT
 
-      # - name: Push container
-      #   working-directory: ./packages/titus-backend
-      #   run: docker push $DOCKER_IMAGE
+      - name: Push container
+        working-directory: ./packages/titus-backend
+        run: docker push $DOCKER_IMAGE
 
-      # - name: Run service
-      #   uses: google-github-actions/deploy-cloudrun@v0.7.0
-      #   with:
-      #     service: ${{ env.CLOUDRUN_SERVICE_NAME }}
-      #     image: ${{ env.DOCKER_IMAGE }}
-      #     credentials: ${{ secrets.GCP_SA_KEY }}
-      #     region: ${{ env.CLOUDRUN_SERVICE_REGION }}
-      #     project_id: ${{ env.PROJECT_ID }}
+      - name: Run service
+        uses: google-github-actions/deploy-cloudrun@v0.7.0
+        with:
+          service: ${{ env.CLOUDRUN_SERVICE_NAME }}
+          image: ${{ env.DOCKER_IMAGE }}
+          credentials: ${{ secrets.GCP_SA_KEY }}
+          region: ${{ env.CLOUDRUN_SERVICE_REGION }}
+          project_id: ${{ env.PROJECT_ID }}
 
       - name: Delete Docker artifacts
         uses: geekyeggo/delete-artifact@v1
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -54,7 +54,7 @@ jobs:
 
       - name: Debug envvar
         run: |
-          echo "The DOCKER_SAVE_ARTIFACT envvar value is (method 1): ${{env.DOCKER_SAVE_ARTIFACT}}"
+          echo "The DOCKER_SAVE_ARTIFACT envvar value is (method 1): ${{ env.DOCKER_SAVE_ARTIFACT }}"
           echo "The DOCKER_SAVE_ARTIFACT envvar value is (method 2): $DOCKER_SAVE_ARTIFACT"
 
       - name: Export Docker image
@@ -65,8 +65,8 @@ jobs:
       - name: Store Docker image as artifact
         uses: actions/upload-artifact@v2
         with:
-          name: $DOCKER_SAVE_ARTIFACT
-          path: /tmp/$DOCKER_SAVE_ARTIFACT
+          name: ${{ env.DOCKER_SAVE_ARTIFACT }}
+          path: /tmp/${{ env.DOCKER_SAVE_ARTIFACT }}
           retention-days: 1
 
   docker-scan:
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - feature/optimize-docker-image
+      - feature/optimize-docker-images
     paths:
       - 'packages/titus-backend/**'
       - '.github/workflows/backend-cd-gcp.yml'
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -53,20 +53,20 @@ jobs:
         run: docker build -t $DOCKER_IMAGE .
 
       - name: Export Docker image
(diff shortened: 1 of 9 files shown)


.github/workflows/backend-cd-gcp.yml
```diff
@@ -44,15 +44,26 @@ jobs:
 
           echo "DOCKER_IMAGE=$DOCKER_REGISTRY_HOST/$PROJECT_ID/titus/backend:$GITHUB_SHA" >> $GITHUB_ENV
 
-      - name: Build container
+      - name: Install Grype tool (vulnerability scan)
+        run: curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin
```
simoneb, line 48: is this the recommended way to install grype? asking because we would usually avoid running anything that's on master, we tend to prefer using something that's published so upgrading to a specific version is an intentional thing, rather than using "whatever is latest"

.github/workflows/backend-cd-gcp.yml
```diff
@@ -44,15 +44,26 @@ jobs:
...
+
+      - name: Scan image for vulnerabilities
+        run: /usr/local/bin/grype $DOCKER_IMAGE | tee /tmp/vuln-scan-titus-backend.txt
+
+      - name: Publish image scan results
+        uses: actions/upload-artifact@v2
+        with:
+          name: vuln-scan-titus-backend.txt
+          path: /tmp/vuln-scan-titus-backend.txt
+          retention-days: 15
```
simoneb, line 62: because installation, execution and publishing of the artifact is something we're doing over and over, why don't we extract that into a reusable github workflow? see https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

packages/titus-backend/Dockerfile
```diff
@@ -1,15 +1,11 @@
-FROM node:16-alpine as builder
-RUN apk add dumb-init
+FROM node:16.13.1-alpine3.14 as builder
```
simoneb, line 1: as an exception to my earlier statement, unless we have an automated mechanism to keep this up to date I wouldn't want us to hardcode a specific version. this is because we are usually ok to use a major version here. node 16, whatever is the latest. I would revert it to the earlier image name. I would still put a dependency mechanism in place as per #1467

packages/titus-backend/Dockerfile
```diff
@@ -1,15 +1,11 @@
...
-COPY --chown=node:node index.js ./index.js
-
-RUN npm ci --only=production
-
+COPY --chown=node:node package.json package-lock.json index.js ./
+COPY --chown=node:node lib lib
+RUN apk upgrade --available && \
+    apk add dumb-init && \
+    rm -rf /var/cache/apk/* && \
+    npm ci --only=production
```
simoneb, line 9: can we keep unrelated commands in different run instructions please? so we don't mix them up

packages/titus-db-manager/Dockerfile
```diff
@@ -1,19 +1,14 @@
...
-COPY --chown=node:node package.json package-lock.json ./
-COPY --chown=node:node lib /titus-db-manager/lib
-
-RUN npm ci --only=production
-
+COPY --chown=node:node package.json package-lock.json index.js migration-start.js ./
+COPY --chown=node:node lib lib
+COPY --chown=node:node seed seed
+COPY --chown=node:node migrate migrate
+COPY --chown=node:node truncate truncate
```
simoneb, line 8: is there a smarter way to do this? it feels overspecified (I do realize it was already like this before)

packages/titus-db-manager/Dockerfile
```diff
@@ -1,19 +1,14 @@
...
+RUN apk upgrade --available && \
+    apk add dumb-init && \
+    rm -rf /var/cache/apk/* && \
+    npm ci --only=production
```
simoneb, line 12: same as earlier comment

simoneb (commented):
See inline comments

simoneb:
Can you please link the issue so when this PR is merged the issue is closed?

packages/titus-backend/Dockerfile
mablanco: Pros and cons already discussed in #1467. As a result, the generic tag is going to be used.

.github/workflows/backend-cd-gcp.yml
mablanco: I'll have a look at it and see what can be done.

.github/workflows/backend-cd-gcp.yml
mablanco: Yes, that's its recommended way (https://github.com/anchore/grype#recommended). Grype checks for the existence of a new tool version and then tries to update its vulnerability db in each execution before doing anything else. It's in its nature to be dynamic and fits perfectly in an ephemeral environment like a Github workflow.

Anyway, it's possible to pin a release version, but I wouldn't recommended it. I think it's better to have the latest available version so we get the full capabilities of Grype in each workflow execution.

packages/titus-backend/Dockerfile
mablanco: Keeping image layers to a minimum is one of Docker's best practices as a way of reducing their size and improving their handling. As a drawback, I agree in that it goes against readability. It's just a matter of choosing one way or the other. What do you prefer?

simoneb: agreed, nonetheless I think we can accept the compromise to slightly increase readability here, but I don't have a strong opinion either

packages/titus-db-manager/Dockerfile
mablanco: It's the expected behaviour of the COPY command when working with directories (this doesn't happen with just files, as you can see in the first line). Trying to reduce the COPY layers lead us inevitably to some kind of workaround, like the one I describe below.

If you want to copy multiple directories (not their contents) under a destination directory in a single command, you'll need to set up the build context so that your source directories are under a common parent and then COPY that parent. So we'd have to move the three folders under one parent folder and add the parent. That may have a huge impact in the source code or even be impossible to do.

Other dirty hacks would go through "tarring" the directories and COPYing the tar file, using several FROM directives or (ugh) playing with the Unix `cp` command. As you can see, overengineering...

.github/workflows/backend-cd-gcp.yml
simoneb: I trust your advice on this but I should point out that not too long ago a tool for code coverage called codecov had a vulnerability for this specific reason. running code in the master branch of any repository is not really a very safe approach because it's "too close" to any potentially harmful activity of an attacker. one commit from a malicious user and we're screwed because we're picking it up right away. instead if we rely on a "released" version of the tool, there are most steps  that an attacker would have to take in order to get malicious code into user's code. if there is a way to use a specific version but loosely, as we're doing with Node 16, I would prefer it

mablanco:
> Can you please link the issue so when this PR is merged the issue is closed?

Done!

packages/titus-db-manager/Dockerfile
simoneb: what about using dockerignore files?

.github/workflows/backend-cd-gcp.yml
simoneb: let's do this also as a learning experience

.github/workflows/backend-cd-gcp.yml
mablanco: Grype's installation is not using the master branch HEAD but the latest release available at GitHub. If you're curious about this, have a look at the installation script and look for the `tag_to_version` function. That's why I rely on its recommended installation. I totally agree with you about the dangers of using an arbitrary commit to install a tool.

packages/titus-backend/Dockerfile
mablanco: OK, let's go for the readability path then.

packages/titus-db-manager/Dockerfile
mablanco: As stated before, we'll go for the readability path.

packages/titus-db-manager/Dockerfile
mablanco: I wouldn't use a static list of files to be ignored from a massive COPY command. If a dev adds files/folders or modifies the repo filesystem, we could end with unwanted files/folders in the image.

.github/workflows/backend-cd-gcp.yml
simoneb: that's a good thing but nonetheless the entry point is an executable script on the main branch. if an attacker manages to push a commit there we're screwed. see https://about.codecov.io/security-update/

mablanco: Do you mean the `install.sh` script? In that case, even pinning the release version wouldn't help as we'd be using the very same installation script. Therefore, to be "completely" safe we should build our own installation procedure (downloading the proper binary from GitHub) or go through the Docker option (but I haven't used it in Github workflows, so I don't know if that would work).

simoneb: One option that I would be more comfortable with is run it from docker instead. They publish images there, even though they don't have documentation. See https://github.com/anchore/grype/issues/388

packages/titus-db-manager/Dockerfile
simoneb: fair enough

.github/workflows/backend-cd-gcp.yml
mablanco: Yes, I'm aware they're quite lacking in that aspect.

Let me have a look at the Docker option and see what I can come up with.

simoneb: It's not a big deal really, but I must say I find it somewhat ironic that a tool for security scanning has such an obvious flaw. All it takes is a malicious commits to screw up users.

.github/workflows/backend-cd-gcp.yml
mablanco: First working version of the reusable workflow for Grype: https://github.com/nearform/titus/actions/runs/1653415544

mablanco: Now I have to add this to the rest of workflows that build Docker images.

.github/workflows/backend-cd-gcp.yml
mablanco: I've tested the Docker image and, while it works great in my local environment from the CLI, it fails when run from a workflow (it doesn't seem to be compatible with GitHub's workflows requirements): https://github.com/nearform/titus/runs/4701694131?check_suite_focus=true

Using the Grype Docker container would also imply to fiddle with volumes to be able to access the Docker tar file and export the vulnerability report, as well as running certain actions inside a container that may not be ready for the task.

As building our own Grype image would take such an investment of effort and time, I consider downloading a specific binary a more profitable path to explore, which is what I'm starting now.

simoneb: that's ok, let's stick with the shell script then. it was an interesting learning opportunity anyway!

mablanco: I'm not using the installation script anymore (even if they have protected their `main branch). I'm working on issuing several sequential commands so a specific release archive is downloaded and extracted and the binary installed in `/usr/local/bin`.

mablanco: I've learned about using Docker containers in a GitHub workflow, though.

.github/workflows/backend-cd-gcp.yml
mablanco: Done! All the CD workflows have run successfully, except for the `storybook` one, which has an unrelated pending issue.

- Backend: https://github.com/nearform/titus/actions/runs/1653640186
- DB Manager: https://github.com/nearform/titus/actions/runs/1653640197
- Frontend: https://github.com/nearform/titus/actions/runs/1653640185

.github/workflows/backend-cd-gcp.yml
mablanco: Thinking about it, I don't need the installation script's extra features, like autodetecting the target platform (arch, OS, etc.). The workflow uses `ubuntu-latest` as runner, so we can go safely with the `linux-amd64` binary.

simoneb (approved):
LGTM, very nice job. Feel free to merge this, then please follow the execution of the workflows on master to make sure that everything works fine (with the exception of storybook)

mablanco:
Thanks @simoneb!

PR #1469, opened 2022-01-04
Fix cloudrun service name

mablanco:
Fix the Cloudrun Service Name in 3 workflows.


mablanco:
I'd like to do some more testing. Closing PR.

Issue #1475, opened 2022-01-10
Unknown IP whitelisted in SG


zimny:
https://github.com/nearform/titus/blob/master/infra/aws/terraform/backend.tf#L254 creates and inbound sg rule that is opening port 80 for 1.1.1.1 IP which seems pretty arbitrary. NLB normally has no static IP like that, the task is to investigate the purpose of this rule and potentially remove it.

mablanco:
1.1.1.1 is a free DNS service by Cloudflare. The service functions as a recursive name server providing domain name resolution for any host on the Internet. Therefore, it should generate only answers to DNS requests, not HTTP requests to our NLB.

I'll investigate further to be sure of the purpose of the rule.

mablanco:
I've checked the specific commit that added this block of code (https://github.com/nearform/titus/pull/983/commits/4a0c9c832d4ca7b981c042203adefd4899d197ea) and there's nothing that helps explain the reasons behind this rule, except for a comment that says "Network interface IPs of NLB". It seems that the purpose of the rule was to open HTTP traffic to the world.

mablanco:
The `cidr_blocks` parameter in a `aws_security_group_rule` block refers to either the source (`ingress`) or the destination (`egress`) of the traffic managed by the rule. Therefore, I think that we can definitely assume the purpose of this rule to be to accept traffic only from the NLB.

I'm going to modify the `cidr_blocks` parameter to reflect the NLB IPs and to delete the `from_port` one (it makes no sense).

mablanco:
I fear it isn't possible to modify the `cidr_blocks` parameter to reflect the NLB IP, as this is dynamically assigned and unknown to Terraform. Besides, Network Load Balancers don't have SGs. More on this issue in https://github.com/hashicorp/terraform-provider-aws/pull/2901#issue-286976429

As the NLB is already connected to the VPC through a LB target group, the LB is attached to 2 private subnets and the ECS instance in deployed inside the private VPC, I think that we could safely assign the `0.0.0.0` value to the `cidr_blocks` parameter. If we wanted to assign specific source CIDR blocks to the SG rule, we'd need an important refactor of the `aws_lb` definition (like this one: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb#specifying-private-ip-addresses-for-an-internal-facing-load-balancer), managing the internal network interface of the NLB and assigning a static private IP to it. This may be too much work for the desired result.

zimny:
[...]

PR #1478, opened 2022-01-10
Open traffic to HTTP port and remove wrong source IP

mablanco:
Please, have a look at discussion at #1475.

### DIFF
--- a/infra/aws/terraform/backend.tf
+++ b/infra/aws/terraform/backend.tf
@@ -251,7 +251,7 @@ resource "aws_security_group_rule" "inbound" {
   from_port   = 80
   to_port     = 80
   protocol    = "tcp"
-  cidr_blocks = ["1.1.1.1/32"] # Network interface IPs of NLB
+  cidr_blocks = ["0.0.0.0/0"] # This could also be the network interface IP of the NLB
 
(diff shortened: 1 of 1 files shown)


infra/aws/terraform/backend.tf
```diff
@@ -251,7 +251,7 @@ resource "aws_security_group_rule" "inbound" {
   from_port   = 80
   to_port     = 80
   protocol    = "tcp"
-  cidr_blocks = ["1.1.1.1/32"] # Network interface IPs of NLB
+  cidr_blocks = ["0.0.0.0/0"] # This could also be the network interface IP of the NLB
```
zimny, line 254: I'd drop the comment about what it could be, and explain that we have NLB in use, not ALB so we can't restrict access to particular IP. What do you think?

mablanco: It's already stated at the beginning of that file, but it doesn't hurt to explain it anywhere we may see fit. What do you propose?

Although I don't like comments, this particular one helped me understand the real purpose of the rule.

mablanco: Great, I like it!

zimny (approved):
LGTM

Issue #1474, opened 2022-01-10
Unknown bucket name hardcoded in aws/terraform


zimny:
https://github.com/nearform/titus/blob/master/infra/aws/terraform/cdn.tf#L42 you can find reference to a bucket called `vittle.ai` which is not defined in the infra code.

Ideally CloudFront should reference with OAI (https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html) bucket created within the terraform code.

mablanco:
I've checked that the S3 bucket defined in the `storage.tf` file is not being used elsewhere in the code. According to the infra diagram (https://github.com/nearform/titus/blob/master/docs/devops/aws/infra.png), there should be one and only one S3 bucket, whose purpose is to serve static files when Cloudfront asks for them. Therefore, the S3 bucket defined in the `storage.tf` file and the one referenced in the `cdn.tf` file should be the same.

Therefore, I'm going to reconcile both files so Cloudfront uses the S3 bucket.

mablanco:
I've connected the CDN distribution with the S3 bucket and restricted access to its content by using an OAI.

mablanco:
Reopening issue to deploy infra and check it live.

mablanco:
Comments in PR #1479.

Issue #1476, opened 2022-01-10
Cloudfront is pointing origin into ELB instead of API gateway


zimny:
https://github.com/nearform/titus/blob/master/infra/aws/terraform/cdn.tf#L26 in here cloudfront is pointing one of the origins into ELB, this is not like in the diagram. I'm confused on what is desired behaviour. 

The task is to first identify desired flow for the dynamic requests and update either diagram or the code to reflect it.

mablanco:
After checking the code, I can confirm that the chain API Gateway > VPC PrivateLink > NLB is correctly defined.

mablanco:
The `domain_name` attribute in the API origin block needs to point to the invoke URL of the API Gateway deployment.

mablanco:
Solved in commit b756c90

PR #1479, opened 2022-01-10
Connect CDN to correct S3 bucket & restrict access by using an OAI

mablanco:
Please read the discussion at #1474.

### DIFF
--- a/infra/aws/terraform/cdn.tf
+++ b/infra/aws/terraform/cdn.tf
@@ -1,3 +1,11 @@
+resource "aws_cloudfront_origin_access_identity" "this" {
+  comment = "Static content"
+}
+
+locals {
+  s3_origin_id = "static-content"
+}
+
 resource "aws_cloudfront_distribution" "this" {
   default_cache_behavior {
     allowed_methods = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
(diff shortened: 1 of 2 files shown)


mablanco:
These latest commits solve the following issues:

- Blocked all public access to S3 bucket (we're using OAI now).
- Atomatically update bucket policy to grant access to the OAI.
- Remove the `origin_path` attributes from both origins; combined with the `path_pattern` attribute (necessary for choosing origin per request) makes Cloudfront expect nested directories in the origins.
- Make the S3 origin the default one and removed its `path_pattern` attribute. Now you can save your static website wherever you want inside the bucket.
- Fix the value of the `path_pattern` attribute for the API GW origin.
- Connect APIGW origin to correct URL instead of ELB (#1476).
- Improve readability by reordering attributes and blocks.

infra/aws/terraform/cdn.tf
```diff
@@ -38,48 +38,52 @@ resource "aws_cloudfront_distribution" "this" {
...
-
   restrictions {
     geo_restriction {
       restriction_type = "none"
     }
   }
+
+  default_cache_behavior {
+    allowed_methods        = ["GET", "HEAD"]
+    cache_policy_id        = "658327ea-f89d-4fab-a63d-7e88639e58f6"
```
zimny, line 49: I suppose this is ID of some managed policy, can you use the datasource instead?
eg
```
data "aws_cloudfront_cache_policy" "this" {
  name = "Managed-CachingOptimized"
}
...
 cache_policy_id          = data.aws_cloudfront_cache_policy.this.id
```
?

mablanco: I'd rather use the `id` as it's not subject to changes as a name can be. However, I understand it's not easy to read and it makes you go to the AWS docs to know about that policy or at least add a comment to state which policy you're using.

What path do you prefer?

zimny: well, magic string is in most cases less preferable than data source :) if possible please use the data source instead

mablanco: ok, I'll go that way

zimny (approved):
LGTM

Issue #1477, opened 2022-01-10
Documentation for getting started seems not correct


zimny:
When I tried to follow the getting started docs for the infra https://github.com/nearform/titus/tree/master/docs/devops/aws I found some information missing. Additionally, it is hard to define whether the code working as expected as there's no landing page available. 

The task is to verify if the getting started docs allow user to get a basic 3 tier web app hosting and if not, report what is the current state.

mablanco:
I will start adding a placeholder `index.html` file in the S3 bucket to make sure we have a working stack just after deployment.

mablanco:
I've been able to upload a static placeholder website (https://github.com/sebastian9486/static-placeholder-website) from TF. Now you get a functional endpoint through Cloudfront when you deploy the infra with Terraform (it's also completely destroyed following the instructions in the README file).

mablanco:
I've also improved the README, fixed a few erratas and checked that the info and the diagram are up to date.

@zimny is this enough to solve this issue or is there anything else required?

PR #1480, opened 2022-01-12
Fix AWS documentation for getting started

mablanco:
Please have a look at #1477


zimny (approved):
For the time being it seems enough. Thanks @mablanco !

PR #1482, opened 2022-01-20
fix: strictly check if NODE_ENV is 'production'

### LINKED ISSUE #1481: Invalid regexes used to check whether the back end and database manager services are running in production mode.
The regexes used to set `config.isProduction` in both services are invalid, always returning `false`, no matter what  value `NODE_ENV` is set to in the respective `.env` file.

Here's the back end code:

https://github.com/nearform/titus/blob/71723b21dacdbd2229f5d6cd0cb10b0924d12e70/packages/titus-backend/lib/config/index.js#L54

And the database manager code:

https://github.com/nearform/titus/blob/743c044c8d7151e098af174f886c8ede141114f6/packages/titus-db-manager/lib/config.js#L22

I'm happy to open a PR to fix this, but I'd like to discuss the solution before I start. Just fix it separately in both services? Since titus is a monorepo project, it could be nice to have a `utils` type module that contains reusable functions, but that seems overkill for this bug.

I think my preference would be to fix it in both services, getting rid of the regexes and either:

1. Being strict about the value - i.e.
`const isProduction = config.NODE_ENV === 'production'`

2. Or else be slightly less strict about the value, e.g.
  `const isProduction = config.NODE_ENV.trim().toLowerCase() === 'production'`

I prefer option # 1, since I'm pretty sure `envSchema` trims any whitespace from environment variables, anyway.

An in-between option (# 3) would be to expect whitespace to be correct, but use `toLowerCase` to be less strict about the capitalization.

hails:
Closes #1481

### DIFF
--- a/packages/titus-backend-typescript/src/config/index.ts
+++ b/packages/titus-backend-typescript/src/config/index.ts
@@ -68,7 +68,7 @@ const routeResponseSchemaOpts = Type.Strict(
   })
 ).properties
 
-const isProduction = /^\s$production\s*$/i.test(config.NODE_ENV)
+const isProduction = config.NODE_ENV === 'production'
 
 // Global configuration, from env variables
 export const configOptions = {
--- a/packages/titus-backend/lib/config/index.js
+++ b/packages/titus-backend/lib/config/index.js
@@ -51,7 +51,7 @@ const routeResponseSchemaOpts = S.object()
   )
   .valueOf().properties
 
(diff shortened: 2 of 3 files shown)


simoneb (approved):
I'm ok with this simple solution. The literal value is what people should be using, so no need to be flexible and accept other values here.

simoneb:
@hails please check with @kurtmilam if he's happy with this change as he's the OP and if he is, let's merge this.

kurtmilam:
LGTM!

PR #1488, opened 2022-02-03
fix: pin chalk dev dependency and run titus-lib test on CI

### LINKED ISSUE #1487: Tests failing in titus-lib after Chalk major version bump
#1387 bumped [Chalk to v5.0.0](https://github.com/chalk/chalk/releases/tag/v5.0.0) in the titus-lib directory.

This package is now ESM only which is causing the unit tests, which are using CJS, to fail (also the tests for this lib do not look to be run as part of the CI).

It's also preventing the config environment scripts from running correctly, e.g. `npm run create:env`.

Example error:

`Error [ERR_REQUIRE_ESM]: require() of ES Module [path redacted]/lib/node_modules/chalk/source/index.js from [path redacted]/lib/tools.js not supported.
Instead change the require of index.js in [path redacted]/lib/tools.js to a dynamic import() which is available in all CommonJS modules.`

Some options for us could be:
- pin Chalk to v4.1.2
- remove it from titus-lib
- use dynamic import to get v5.0.0 working with CJS

Any thoughts on how best to proceed?

Node version: v16.13.2

krisgrint:
Fixes #1487 

This came about when getting Titus to run locally.

I've pinned the dependency and removed it where it was not being used.

I've added a new CI action to ensure the titus-lib tests are run if the directory is modified.

### DIFF
--- a/packages/titus-frontend/package.json
+++ b/packages/titus-frontend/package.json
@@ -54,7 +54,6 @@
     "@storybook/react": "^6.4.9",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.0.0",
-    "chalk": "^5.0.0",
     "dotenv": "^10.0.0",
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.3.0",
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,6 +7,7 @@ updates:
     ignore:
       - dependency-name: 'aws-cdk'
       - dependency-name: '@aws-cdk/*'
+      - dependency-name: 'chalk'
   - package-ecosystem: github-actions
     directory: '/'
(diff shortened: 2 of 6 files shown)


packages/titus-frontend/package.json
```diff
@@ -54,7 +54,6 @@
     "@storybook/react": "^6.4.9",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.0.0",
-    "chalk": "^5.0.0",
```
simoneb, line 57: is this not needed?

krisgrint: I can't find it used directly in the frontend project. It is used by some of our devDependencies (like Jest) but at a lower version.

`depcheck` also flags it as unused.

simoneb (approved):

PR #1515, opened 2022-04-13
fix: detect production environment

### LINKED ISSUE #1484: Devise a new mechanism to detect in which environment the application is running
Currently, we're relying on `NODE_ENV` to detect in which environment a titus application is running. 
For example, I've recently introduced a [change](https://github.com/nearform/titus/pull/1482) on how we detect if the application is running in a production environment.
One approach would be setting a custom environment variable instead.

fahric:
Closes #1484 
add a new env variable to detect live env

### DIFF
--- a/packages/titus-backend-typescript/.env.sample
+++ b/packages/titus-backend-typescript/.env.sample
@@ -1,4 +1,5 @@
 NODE_ENV=development
+IS_PRODUCTION=false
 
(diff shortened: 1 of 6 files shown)


simoneb:
@fahric any reason why this is still draft?

kurtmilam (approved):
lgtm!

simoneb (approved):

PR #1518, opened 2022-04-26
adding workflow_dispatch to have the ability to trigger the workflow from the UI

### LINKED ISSUE #1517: TITUS - Add a manual trigger to workflow Backend CD to GCP
In some situations we may want to trigger the workflow manually, for example with a change at Terraform files this workflow will not be triggered automatically which is expected, nevertheless we still want to trigger this workflow to be sure that everything is working as expected.

ital0sant0s:
This fix #1517 and it's adding workflow_dispatch to have the ability to trigger the workflow from the UI

### DIFF
--- a/.github/workflows/backend-cd-gcp.yml
+++ b/.github/workflows/backend-cd-gcp.yml
@@ -1,6 +1,8 @@
 name: 'Backend CD to GCP'
 
 on:
+  workflow_dispatch:
+  
   push:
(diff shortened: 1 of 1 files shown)


simoneb (commented):
Shall we do the same to the other CD workflow for consistency please?

williamwd (approved):

PR #1526, opened 2022-06-17
Update truncate logic

toni-sharpe:
Truncate needs CASCADE to avoid "cannot truncate a table referenced in a foreign key constraint" errors

Following this: https://stackoverflow.com/a/56700574

Fixes https://github.com/nearform/titus/issues/1534

### DIFF
--- a/packages/titus-db-manager/truncate/index.test.js
+++ b/packages/titus-db-manager/truncate/index.test.js
@@ -15,7 +15,7 @@ describe('truncate', () => {
     expect(pg.query).toHaveBeenCalledWith(`
     TRUNCATE TABLE
       foo.some_table
-    RESTART IDENTITY;
+    RESTART IDENTITY CASCADE;
   `)
   })
 })


packages/titus-db-manager/truncate/index.js
```diff
@@ -4,7 +4,7 @@ module.exports = async function truncate(pg, { schema, logger }) {
   await pg.query(`
     TRUNCATE TABLE
       ${schema}.some_table
-    RESTART IDENTITY;
+    RESTART IDENTITY CAASCADE;
```
simoneb, line 7:
```suggestion
    RESTART IDENTITY CASCADE;
```

simoneb:
@toni-sharpe do you have any capacity to fix the failing tests? it should be fairly straightforward

simoneb:
@toni-sharpe ping

toni-sharpe:
@simoneb I did the tests and the build is green here now, thanks for the ping!

(I'll be away 'til 01.08 after today)

simoneb (approved):

